PR #52, opened 2021-11-04
chore: address retroactive stability increases in versioning doc

cmars:
Amend the versioning standard to prohibit retroactive stability
increases on existing versions, with an example of why it causes
problems.


docs/version.md
```diff
@@ -67,13 +63,25 @@ In increasing order of stability, from least to most stable, these are:
     - During its availability, it must not be revised with incompatible and breaking changes.
     - It may be revised with backwards-compatible, additive changes.
 
-### Increasing stability of a release
+### Increasing stability of a resource over time
```
ianlivingstone, line 66: Another way for us to frame this: promotion of a route requires a new route + the old one to continue to be routed.

Do we have an issue to track implementing a rule?

cmars: Updated the framing, added #53. PTAL

genslein (approved):
Nice catch!

PR #56, opened 2021-11-08
fix: schema properties should use snake case names

cmars:
Per discussion today w/ @ianlivingstone, we want property names to be
consistent with parameter names, which means snake_case.


cmars:
CC @acunniffe

gablaxian (approved):

PR #57, opened 2021-11-09
initial migration to optic-ci & documentation

acunniffe:
[ready for optic-main branch, not ready for primetime] 

Here's the first batch of changes to bring optic-ci into sweater-comb. Things I expect to work:
- tests for your CI checks
- build
- instructions in `DEVELOP.md` for running it yourself. 

Things I do not expect to work:
- circle CI (we'll see once we open it)
- publishing docker or npm
- all your rules (see TODO.md) 

Next steps:
- Optic, finish writing rules
- Together -- build, dockerise, etc
- Snyk, can we get a few scenarios into `/end-end-tests`? I'll write a script to iterate each subfolder and snapshot the test results so we get good gaurentes about stability around our core use cases. The unit tests are great, but a few of these will make me feel very confident 
   - One where there is a from, but not to (deleted spec), with a context that should pass / fail a check
   - One where is a a to, but no from (new date version) 
   - A few examples with breaking changes you hope to catch in the to variant.


DEVELOP.md
```diff
@@ -0,0 +1,32 @@
...
+```bash
+yarn install
+source sourceme.sh
+```
+
+This will install the dependencies, and put `opticCiDev` on your path for the duration of that terminal's lifespan.
+
+Running the compare command:
+```bash
+opticCiDev compare \
```
acunniffe, line 12: @cmars -- here's how to run it yourself!

Issue #68, opened 2021-11-11
[Optic CI] Kebab case rule does not allow single-word names


cmars:
The kebab-case rule on header keys is failing on single-word keys, which should be allowed.

<img width="215" alt="Screen Shot 2021-11-11 at 12 46 06 PM" src="https://user-images.githubusercontent.com/23741/141352271-c72d2b15-54ce-45f9-a073-b012d2844825.png">

To reproduce: `node_modules/.bin/ts-node src/index.ts compare --to end-end-tests/api-standards/resources/thing/2021-11-10/000-baseline.yaml --context '{"changeDate":"2021-11-11","changeResource":"thing","changeVersion":"2021-11-10"}'`

CC @acunniffe

cmars:
Fixed by #72

Issue #67, opened 2021-11-11
[Optic CI] Snake case parameter name rule failing a valid param


cmars:
`thing_id` should be a valid snake case param, but it's getting rejected:

<img width="231" alt="Screen Shot 2021-11-11 at 12 39 51 PM" src="https://user-images.githubusercontent.com/23741/141351631-14b36f77-cf9c-41f7-8b33-ae03c5edede8.png">

To reproduce: `node_modules/.bin/ts-node src/index.ts compare --to end-end-tests/api-standards/resources/thing/2021-11-10/000-baseline.yaml --context '{"changeDate":"2021-11-11","changeResource":"thing","changeVersion":"2021-11-10"}'`

cmars:
CC @acunniffe

cmars:
Fixed by #72

Issue #65, opened 2021-11-11
[Optic CI] Required parameters should be allowed in new versions


cmars:
We don't want to allow required parameters to be added once a version has been introduced -- but we need to be able to introduce them initially.

<img width="232" alt="Screen Shot 2021-11-11 at 12 04 36 PM" src="https://user-images.githubusercontent.com/23741/141347742-9e1ad7cd-27b9-4c04-91bc-e2f31d276482.png">

When running `node_modules/.bin/ts-node src/index.ts compare --to end-end-tests/api-standards/resources/thing/2021-11-10/000-baseline.yaml --context '{"changeDate":"2021-11-11","changeResource":"thing","changeVersion":"2021-11-10"}'`

cmars:
CC @acunniffe

acunniffe:
@cmars -- ahhh this happens when you don't specify a from. It's triggering the changelog rules, even though technically, it's not really being added. We know how to handle this case -- thanks for the test data. Will get this passing

cmars:
Fixed by #72

PR #72, opened 2021-11-14
every issue should have a sourcemap, end-end tests should run & logic fixes

acunniffe:
This is a massive PR (mostly in saved snapshot tests for the scenarios). 

1. @cmars noticed that most of the issues were being logged without links to the file / line where they could be resolved -- that's a pretty annoying UX for a developer (x is broken...good luck). Spent a lot of time getting this calibrated so it worked with all kinds of OpenAPI files / refs. Every end-end test has a sourcemapping now, and you can see it when you run the compare command
Screen Shot 2021-11-14 at 2 38 35 PM
 
2. We added end-end tests (see further comments for info) using Jest / snapshots to replace the bash script. Since they're snapshots all regressions in behavior should be caught. Bonus, if we update the error messages, you'll see them in the snapshots too and can approve new copy :) 

3. Implemented fixes for: 
https://github.com/snyk/sweater-comb/issues/68
https://github.com/snyk/sweater-comb/issues/67
https://github.com/snyk/sweater-comb/issues/65


cmars (approved):
Thanks for this. The output format is better. There are still some rough edges:

* If I redirect output to a file, I still see some wrapping. We'll want machine readable output in followups: JSON and plain text (no formatting, terminal ANSI codes, etc). Let me know when you're ready to implement that and I can provide guidance.
* The same rule failure is repeated many times when it is a problem in one of the $ref'd common schema entities. It would be better if we only write a given instance of a rule failure once. Then the developer is not overwhelmed with a flood of errors on what could be a small set of problems in a heavily referenced entity.

The jest snapshot for the end-end-test scenarios is very large -- about 2mb. Is it necessary to store all the passing checks in there?

I noticed that many of the rule failures are missing examples on the JSON API structures themselves. I can look into that tomorrow.

acunniffe:
> The same rule failure is repeated many times when it is a problem in one of the $ref'd common schema entities. It would be better if we only write a given instance of a rule failure once. Then the developer is not overwhelmed with a flood of errors on what could be a small set of problems in a heavily referenced entity.

I can see the motivation for doing this, and agree, it is too noisy today. Maybe we can do a "Shared Components" section that groups issues by their references?

The only downside to doing so is that some rules might behave differently. Usage context matters a lot for breaking changes -- ie a field added in a Shared Schema is breaking, only in requests, and not in responses. The same property 'added' might be breaking sometimes but not others. That's why our rules treat each usage as its own instance and groups them by the operation where they are used. 

I think a good middle ground that'd calibrate this better is making a distinction between our `added|changed|removed` rule and our `requirement` rule. Requirements are more structural (must have name X), and (must have example), and don't concern themselves with the usage context as much. Further proof this might be a good angle to tackle the problem -- all the noise today comes from `requirement` rules

So (pending your approval), I propose a plan to
- group breaking change rules with their operations (even if there's a few repeats)
- remove duplicates and group all requirements from refs under a "shared components" section in the output 

Happy to discuss more

PR #73, opened 2021-11-15
Optic ci/dockerize

acunniffe:
Added a proper Dockerfile that builds, and can run tests against the examples in end-end tests

```
docker build . -t opticomb:latest
docker run -it opticomb:latest compare --to end-end-tests/api-standards/resources/thing/2021-11-10/000-baseline.yaml --context '{"changeDate":"2021-11-11","changeResource":"thing","changeVersion":"2021-11-10"}'
```

This should support a volume being mounted (just like the sweater comb example https://github.com/snyk/sweater-comb/blob/a7b1169ff5ea233afec086e22a7ce5d336f31e94/Dockerfile#L18) 

Questions for @cmars 
- is this Dockerfile ok? Are there Snyk standards somewhere we need to follow? 
- should we add a volume or leave that to you as you integrate? 
- need any help publishing this Dockerfile to the same place you publish Sweater Comb?


Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:14-alpine
+ENV NODE_ENV=development
```
cmars, line 2: Why is this needed?

Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+WORKDIR /usr/src/app
+COPY ["package.json", "yarn.lock", "./"]
+RUN yarn install
+COPY . .
```
cmars, line 6: Why are there two COPY commands?

cmars (approved):
I think this is OK, just a couple of questions below.

Dockerfile
devdoshi: Since it is installing and building, it needs the development dependencies. If we were consuming a published npm artifact we would use production. This dockerfile can be cleaned up for production by using multi stage builds and stuff but it would be best if it didn't have to do the building

Dockerfile
devdoshi: It's 2 stages, for docker layer caching reasons. If the package json and yarn lock haven't changed, docker doesn't have to rerun those stages, and if the rest of the copied stuff hasn't changed either then that can skip too. Ideally it would be even more granular than copy . . And only copy the stuff needed for building

Dockerfile
devdoshi: It could be set to production in later steps but at least for the install it needs development.

Dockerfile
cmars: Nice! Thanks for the background, I'm still getting familiar with Node & the JS ecosystem.

Issue #79, opened 2021-11-15
Breaking change rule output when an operation is removed


cmars:
This breaking change scenario fails, but the output is not very clear about what needs fixing.

Steps to reproduce:

```
$ node_modules/.bin/ts-node src/index.ts compare \
  --from end-end-tests/api-standards/resources/thing/2021-11-10/000-baseline.yaml \
  --to end-end-tests/api-standards/resources/thing/2021-11-10/001-fail-operation-removed.yaml \
  --context '{"changeDate":"2021-11-11","changeResource":"thing","changeVersion":"2021-11-10"}' 2>&1 | pbcopy
```

Output:

```
Loading specifications for comparison:
Current specification: loading...
Next specification: loading...
[2K[1A[2K[1A[2K[1A[2K[GLoading specifications for comparison:
Current specification: done
Next specification: loading...
[2K[1A[2K[1A[2K[1A[2K[GLoading specifications for comparison:
Current specification: done
Next specification: done
running rules...
[2K[1A[2K[1A[2K[1A[2K[1A[2K[GLoading specifications for comparison:
Current specification: done
Next specification: done


  PASS  DELETE /thing/{thing_id}


  PASS  GET /thing


  PASS  GET /thing/{thing_id}


  FAIL  PATCH /thing/{thing_id}
  x removed response: undefined must not be removed
    Cannot read property 'statusCode' of undefined
  x removed response: undefined must not be removed
    Cannot read property 'statusCode' of undefined
  x removed response: undefined must not be removed
    Cannot read property 'statusCode' of undefined
  x removed response: undefined must not be removed
    Cannot read property 'statusCode' of undefined
  x removed response: undefined must not be removed
    Cannot read property 'statusCode' of undefined
  x removed response: undefined must not be removed
    Cannot read property 'statusCode' of undefined
  x removed response: undefined must not be removed
    Cannot read property 'statusCode' of undefined
  x removed response: undefined must not be removed
    Cannot read property 'statusCode' of undefined
  x removed field: undefined must not be removed
    Cannot read property 'key' of undefined
  x removed field: undefined must not be removed
    Cannot read property 'key' of undefined
  x removed field: undefined must not be removed
    Cannot read property 'key' of undefined
  x removed field: undefined must not be removed
    Cannot read property 'key' of undefined
  x removed field: undefined must not be removed
    Cannot read property 'key' of undefined
  x removed field: undefined must not be removed
    Cannot read property 'key' of undefined
[...]

PR #80, opened 2021-11-17
Fixes for important issues

acunniffe:
This PR lands a ton of fixes that required updates to the core rules engine. 

In a world where you can write added / removed / changed rules, you need to make sure you qualify nested rules before they runs. For instance if you remove a response, the response.properties removed rules should not run, else it's too noisy. If you add an operation, you don't want all the request/response/property/param added rules to run. 

Some core restructuring has taken place to make this behavior automatic and general across all nested entities in the API schema. 

Fixes: 
- https://github.com/snyk/sweater-comb/issues/79

You'll notice a lot of snapshot was deleted in this PR, that's because all the rules running at the wrong times have been fixed.


cmars (approved):
LGTM, thanks for this!

I'm seeing a lot of formatting changes. We probably need prettier linting to reduce these for more readable diffs, something I will look into.

acunniffe:
> We probably need prettier linting to reduce these for more readable diffs, something I will look into.
oh yeah, I don't think we put a config in the root so our computer globals are taking over. Feel free to set one up and I'll make sure our next PRs are made with them running

PR #86, opened 2021-11-17
chore: relocate JSON API components for reuse

cmars:
Moving JSON API types up to the top level for easier reuse.


cmars:
On hold pending #87

Issue #78, opened 2021-11-15
Should not be allowed to change stability


cmars:
This should fail a check, but it currently passes:

```
node_modules/.bin/ts-node src/index.ts compare \
  --from end-end-tests/api-standards/resources/thing/2021-11-10/000-baseline.yaml \
  --to end-end-tests/api-standards/resources/thing/2021-11-10/001-fail-stability-change.yaml \
  --context '{"changeDate":"2021-11-11","changeResource":"thing","changeVersion":"2021-11-10"}'
```

Basically, we do not want the stability level to be changed if the `--from` stability is experimental or higher.

We'll allow `wip` to be changed to a higher level, because `wip` stability is not exposed to users. And in general, `wip` should be exempt from breaking change rules for this reason.

cmars:
CC @acunniffe

cmars:
This has been fixed (thank you!)

PR #87, opened 2021-11-18
CLI output options and issue resolution

acunniffe:
In this PR we've released a few priorities based on today's feedback
- `--output=json` and `--output=plain` can be passed into the compare command. 
- we've tested JSON with a lot of things we know @cmars has tried. piping to a file `... > file.json`, reading it `... | cat` and even with jq `... | jq`. Lots of fun options possible here
- `plain` mode is great for copying around and opening more GH issues :) 

We've also released fixes for: 
- https://github.com/snyk/sweater-comb/issues/79
- https://github.com/snyk/sweater-comb/issues/78


src/rulesets/tests/__snapshots__/end-end.test.ts.snap
```diff
@@ -64137,6 +64219,40 @@ Array [
     },
     "where": "added field: strangeness",
   },
+  Object {
+    "change": null,
+    "condition": "be provided for every resource document",
```
acunniffe, line 64224: @cmars  -- looks like some of the examples are failing now because stability is missing

src/rulesets/tests/__snapshots__/end-end.test.ts.snap
```diff
@@ -23349,6 +23383,20 @@ Array [
...
+    "docsLink": undefined,
+    "error": "expected operation to be present",
+    "isMust": true,
+    "isShould": false,
+    "passed": false,
+    "sourcemap": Object {
+      "filePath": undefined,
+      "preview": "",
+    },
+    "where": "removed operation: PATCH /thing/{thing_id}",
```
acunniffe, line 23398: Fix for operation removal not being detected

src/dsl.ts
```diff
@@ -27,7 +27,7 @@ import { genericEntityRuleImpl } from "@useoptic/api-checks/build/sdk/generic-en
 import { ShouldOrMust } from "@useoptic/api-checks/build/sdk/types";
 import { jsonPointerHelpers } from "@useoptic/json-pointer-helpers";
 
-type SnykStablity = "wip" | "experimental" | "beta" | "ga";
+type SnykStablity = "wip" | "experimental" | "beta" | "ga" | "deprecated";
```
cmars, line 30: `deprecated` is a lifecycle stage; not something an API can declare in its OpenAPI spec with an extension.

src/rulesets/api-lifeycle.ts
```diff
@@ -1,14 +1,38 @@
...
         context.changeVersion.date;
       },
     );
   },
+
+  stabilityRequirement: ({ stability }: SnykApiCheckDsl) => {
+    stability.must(
+      "be provided for every resource document",
+      (before, after, context, docs) => {
+        const allowed = ["wip", "experimental", "beta", "ga", "deprecated"];
```
cmars, line 19: Please remove deprecated, this is not something a resource can declare.

cmars (changes requested):

acunniffe:
> `deprecated` is a lifecycle stage; not something an API can declare in its OpenAPI spec with an extension.

Hmm -- makes sense. I removed it in the latest (building now). 

Out of curiosity, where is this information "set" and where are the dates of deprecation registered? Might this also be something useful to put in the "...API lifecycle database" we talked about on slack / our call? Feel free to reply elsewhere

cmars (approved):

PR #89, opened 2021-11-19
chore: fix example, remove format requirement rule

cmars:
Deciding against requiring formats. Custom formats cause too many
problems with multiple systems that need to process OpenAPI. Requiring
examples satisfies the core documentation needs.

So, the rule requiring formats on all string properties is intentionally
removed.

Improving constraints over free-form string types will require a
different approach.


cmars:
See https://github.com/snyk/vervet/issues/72 for where a better alternative to `format:` is being developed.

acunniffe:
Looks good to me, thanks for updating the snapshots. Made it really easy to review the effect of the code changes :)

PR #99, opened 2021-12-01
Landing remaining rules

acunniffe:
This PR has been a 2 week effort from @smizell to:

- port API Next rules from spectral to readable code
- port Json API rules from spectral to readable code
- make Stoplight's OAS 3 rules work with optic-ci

Let us know how these look, specifically if the API Next and Json API rules feel more readable and maintainable (one of the goals). 

Also noted and commented, the spectral compatibility feels like a big win here. If there are other rulesets you want in the future from the spectral community, they should work OOTB.


src/service.ts
```diff
@@ -26,11 +27,26 @@ export function newSnykApiCheckService() {
...
   snykRulesService.useDslWithNamedRules(
     dslConstructor,
     require("./rulesets/api-lifeycle").rules,
   );
+  snykRulesService.useDslWithNamedRules(
+    dslConstructor,
+    require("./rulesets/specification").rules,
+  );
+
+  snykRulesService.useSpectralRuleset({
```
acunniffe, line 39: How spectral rules can be added

src/rulesets/jsonapi.ts
```diff
@@ -0,0 +1,342 @@
...
+      (response, context, docs, specItem) => {
+        if (isOpenApiPath(context.path)) return;
+
+        // Response data
+        if (
+          ["get", "post"].includes(context.method) &&
+          [200, 201].includes(response.statusCode)
+        ) {
+          const responseSchema =
+            specItem.content?.["application/vnd.api+json"]?.schema;
```
cmars, line 282: If the developer mis-types the content type, uses something else like application/json, is the rule error message meaningful enough to the developer, to understand what needs fixing?

cmars (commented):
First pass comments so far. Really excited to try this out. I believe there are some lint errors, `yarn run format` should fix them up.

src/dsl.ts
```diff
@@ -366,7 +419,7 @@ export class SnykApiCheckDsl implements ApiCheckDsl {
                 method: "",
               },
               kind: "ContextRule",
-            },
+            } as any,
```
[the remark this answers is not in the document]
acunniffe: These are custom rules that don't really come out of OpenAPI (in this example its rules about your Context)

We added typing in the library last week to make the OpenAPI traverser more reliable / type safe, but these changes did not fit in. We need to think about custom rule types and how we want them to interop going forward. Definitely temporary

src/rulesets/properties.ts
```diff
@@ -1,44 +1,108 @@
...
-    });
+  propertyFormat: ({ bodyProperties }: SnykApiCheckDsl) => {
+    bodyProperties.requirement.should(
+      "have a format when a string",
+      ({ flatSchema }, context) => {
+        if (flatSchema.type !== "string") return;
+        if (!withinAttributes(context)) return;
+        expect(flatSchema.format).to.exist;
+      },
+    );
```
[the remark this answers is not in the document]
acunniffe: Makes sense, you're right this stuff is hard to enforce across languages. 

Do you want us to comment this out, remove it?

src/rulesets/jsonapi.ts
acunniffe: It made sense in our testing, but it's hard to say what the average developer will know about the guidelines. Do you have an idea what you'd want it to say? 

Also we should go through and add a docsLink for each rule that we can so developers can click into sweater comb docs

acunniffe:
> First pass comments so far. Really excited to try this out. I believe there are some lint errors, `yarn run format` should fix them up.

Sorry about that. All our internal repos have single quote on, so when we copy code in and out it gets messed up. I realized we should be running `format` and will do so going forward!

src/rulesets/jsonapi.ts
cmars: > It made sense in our testing, but it's hard to say what the average developer will know about the guidelines. Do you have an idea what you'd want it to say?

As I'm writing this, I realize this might be better expressed as a different rule (all 2xx responses except 204 need an `application/vnd.api+json` response), but seeing it referenced here made me wonder how it'd be handled.

> Also we should go through and add a docsLink for each rule that we can so developers can click into sweater comb docs

Yes please, that would be great!

src/rulesets/properties.ts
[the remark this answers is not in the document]
cmars: Let's remove it please.

src/rulesets/jsonapi.ts
cmars: Looks like the 2xx rule I'm thinking of is already here: https://github.com/snyk/sweater-comb/pull/99/files#diff-efdbbec0a6118d1d560fdec0ba5b89c1cf08991e76c3f6d8c5143f4229c6c22dR97.

cmars:
I'd be OK with single quote (less typing!) -- as long as it's all consistently formatted. Though, I'd rather do that change in a separate PR, to keep the diffs readable.

cmars:
Just one more thing, can we update json-schema and jsonpointer to address these vulns?

Screen Shot 2021-12-02 at 10 50 40 AM

Screen Shot 2021-12-02 at 10 48 36 AM

acunniffe:
These appear to be in `stoplight/spectral-ruleset` so I can't update them w/o forking and upstreaming. Looks like you use `"^1.2.2"` in `main`. That appears vulnerable too.

cmars:
Thanks for checking! We'll have to ignore them for now and revisit later. For a CI tool, the risk is low, but a service that uses this may need more careful consideration down the road.

cmars (approved):

PR #112, opened 2021-12-08
Lifecycle rules

smizell:
This adds a rule for ensuring the sunset rules are followed. 

* If there is no `deprecatedBy`, then it fails saying it should be deprecated first
* If there is a `deprecatedBy`, it will check the dates to make sure it's been long enough


src/rulesets/api-lifeycle.ts
```diff
@@ -35,4 +35,45 @@ export const rules = {
...
+        (contextDate.getTime() - resourceDate.getTime()) /
+        (1000 * 60 * 60 * 24);
+
+      // Number of days required before a resource can be removed
+      const sunsetSchedule = {
+        experimental: 30,
+        beta: 90,
+        ga: 180,
+      };
+      const requiredDays = sunsetSchedule[stability];
```
cmars, line 70: This rule should ignore `wip` stability.

src/dsl.ts
```diff
@@ -421,13 +424,22 @@ export class SnykApiCheckDsl implements ApiCheckDsl {
...
+              syntheticChange,
+              docsHelper,
+              "api lifeycle: ",
+              statement,
+              must,
+              () =>
+                handler(
+                  {
+                    ...this.providedContext,
+                    wasDeleted: this.nextJsonLike?.info?.title === "Empty",
```
cmars, line 438: Where does "Empty" come from?

src/rulesets/tests/lifecycle.test.ts
```diff
@@ -101,4 +101,65 @@ describe("lifecycle", () => {
...
+              },
+            },
+          },
+        },
+      };
+
+      it("fails when the schedule isn't met", async () => {
+        const result = await compare(baseOpenAPI)
+          .to((spec) => {
+            spec.info.title = "Empty";
```
cmars, line 142: What if I compare a non-empty OpenAPI spec that happens to have the info.title "Empty"? Is there a more robust way to express this?

src/rulesets/api-lifeycle.ts
smizell: Ah, the entire rule, correct? So you can delete a wip even if it's not deprecated?

cmars: Yes, that is fine. `wip` releases are never published.

src/rulesets/tests/lifecycle.test.ts
cmars: Per side-channel discussion, a special-use extension would be best here, to make a precise distinction between content and processing metadata. So maybe `spec.extensions['x-optic-ci-empty-spec'] = true` or something to that effect? We can then lay claim to all extensions `x-optic-ci-*`.

While I'll admit an API called "Empty" is really unlikely (or is it... maybe it's a placeholder name used for a generated template OpenAPI, a mock, or "empty.io", the service that stares back when you stare into the void for too long!), for me it's more about providing a precise contract & behavior.

acunniffe: I'm all for it, @smizell gave me write permissions on this PR, I'll make the change today

acunniffe: waiting on a library release to get this out. We're also bundling some performance fixes for bulk-mode and compare.

acunniffe:
Resolved conflicts (sorry hadn't seen that pop up). 

We're going to fix the `spec.extensions['x-optic-ci-empty-spec'] = true` w/ the next library rollout. This should be good to go

src/rulesets/api-lifeycle.ts
cmars: I think this still needs to be dealt with. If we land as-is, we won't be able to remove `wip` versions? Or the behavior will be undefined (what does "(number) < null" evaluate to in JS?)

src/rulesets/api-lifeycle.ts
```diff
@@ -41,4 +41,46 @@ export const rules = {
       },
     );
   },
+  followSunsetRules: ({ checkApiContext }: SnykApiCheckDsl) => {
+    checkApiContext.must("follow sunset rules", (context) => {
+      if (context.changeVersion.stability === "wip") return;
```
acunniffe, line 46: @cmars  -- see change

PR #115, opened 2021-12-10
chore: demonstrate issue w/failing rules and changes

cmars:
In order to be able to introduce Optic into a project, it should
only apply rules to changes between specs. This allows Optic to be
introduced into a new project which may not yet conform to all the
standards. New changes to specs must meet these standards, but existing
specs should not fail CI.

This change introduces a spec that violates naming conventions, which
should fail if introduced as new (--to 000-fail-naming.yaml with no
--from) but should pass if compared against itself (no change means no
violating change).


cmars:
Landed in rebased #126

Issue #116, opened 2021-12-10
Rules fail even if there is no change that violates them; unexpected behavior


cmars:
If a rule (typically naming) is violated in both the from and to, Optic CI fails. The problem is, we have some API specs that landed prior to some of our standards being complete, and/or completely implemented in the Spectral rules.

Optic CI can be introduced into such a project if it applies rules to the _changes between --from and --to_; that would have the nice “forcing function” of requiring a team to release a new API version once they want to significantly change a non-compliant one. And, it’d be fairly non-blocking — if they complained, we could always say, “hey, no problem, cut a new version and do what you need to do!“.

To address the special case of "let's run all the rules on everything to see where we're at wrt conformance" we could then tell Optic CI "everything is new". But the normal CI case should be "only apply rules to spec changes in this PR".

To reproduce: See #115. Using fixture introduced there:

`optic-ci compare --from 000-fail-naming.yaml  --to 000-fail-naming.yaml` should PASS and exit 0, because that is effectively "no change". Even though the spec introduced from scratch with `optic-ci compare --to 000-fail-naming.yaml` should FAIL.

acunniffe:
Ok the rules this would happen with are all the `requirement` kinds (mostly naming)

Essentially the functionality you're asking for is a flag that would disable, mask, or change the priority from `must` to `should` for requirement rules that fail in both the `from` spec and the `to` spec. 

- (typical case) If they fail in only `to` spec, they're new issues that must be addressed before commiting
- If they fail in only the `from` spec, they're old issues that are now not showing up in `to`. A fixed requirement failure. It might be fixed in a way that introducing breaks along the way -- so some issues might be traded for another. Ie renaming a field to match the naming rule, is a breaking change is some cases. One issue goes away, another appears. 
- (this new case you're asking for) If they fail in both `from` and `to` they were there before optic was introduced (presumably) or someone said it was ok to skip the requirment, and carrying on into new versions. 


an appropriately named flag for this behavior would be something like
```
--allow-from-spec-requirements-to-fail
```

I wouldn't recommend leaving something like this on all the time though. Maybe just until all the current spec versions are bumped to your new standards. 

This is generally a good story to start thinking about rule lifecycle more broadly.

Does something like this make sense?

cmars:
I don't consider this a new case I'm asking for, and I'm not asking for a new flag.

I consider this to be the core functionality and key differentiator of Optic over other solutions like Spectral, and the expected behavior.

[...]

PR #120, opened 2021-12-13
Tune rules w/ more advanced changelog selectors

acunniffe:
As discussed in #116, some rules should only apply when parts of the spec are added, changed, or removed. There are other classes of rules, requirements, that must hold all the time (not matter the history), and when something is changed.

With these 5 selectors we can express everything we currently need. The issue with #116 is that we were using `requirement` in places where we should have been using `added` or `requirementOnChange` 

- This PR tunes the rules in accordance to the choices made with @cmars in a sync call this AM. 
- Uses a stable version of the libraries (demo PRs were on prerelease w/o thorough testing or stability guarantees) 
- We made `added` and `changed` API compatible with `requirement` -- so it's easy to swap selectors w/o having to rewrite your check's handlers, or reorder your function arguments. 

Re: point 3 -- if we need to further change these rules, the fix should be easy 👍 

Really happy with how this turned out and how many things we figured out in the process of triaging this since Friday. @cmars over to you. Snapshots are pretty hard to review here because so many of the rules have disabled themselves and only qualify themselves on change now.


src/rulesets/operations.ts
```diff
@@ -47,6 +47,15 @@ export const rules = {
...
   },
+  operationIdSet: ({ operations }: SnykApiCheckDsl) => {
+    operations.requirement.must(
+      "have operationId",
+      (operation, context, docs) => {
+        docs.includeDocsLink(links.standards.operationIds);
+        expect(operation.operationId).to.be.ok;
+      },
+    );
+  },
```
acunniffe, line 58: require an operationId, always
compliments rule that checks if it's correct on change

src/rulesets/operations.ts
```diff
@@ -73,7 +82,7 @@ export const rules = {
     );
   },
   parameterCase: ({ operations }: SnykApiCheckDsl) => {
-    operations.requirement.must(
+    operations.requirementOnChange.must(
```
acunniffe, line 85: checks if operationId is valid only when it is changed

src/rulesets/tests/operations.test.ts
```diff
@@ -357,22 +357,5 @@ describe("operation parameters", () => {
...
-              name: "version",
-            },
-          ];
-          return spec;
-        })
-        .withRule(rules.parameterCase, emptyContext);
-
-      expect(result.results[0].passed).toBeTruthy();
-      expect(result).toMatchSnapshot();
-    });
```
acunniffe, line 376: this test was never testing what it said it was.

src/rulesets/tests/operations.test.ts
```diff
@@ -46,7 +46,7 @@ describe("operationId", () => {
           delete spec.paths!["/example"]!.get!.operationId;
           return spec;
         })
-        .withRule(rules.operationId, emptyContext);
+        .withRule(rules.operationIdSet, emptyContext);
```
acunniffe, line 50: was running with the wrong rule

src/rulesets/operations.ts
```diff
@@ -31,7 +31,7 @@ const preventParameterChange = (schemaProperty: string) => {
 
 export const rules = {
   operationId: ({ operations }: SnykApiCheckDsl) => {
-    operations.requirement.must(
+    operations.requirementOnChange.must(
       "have the correct operationId format",
```
acunniffe, line 35: only check if operations have the right format on change

cmars (approved):

PR #127, opened 2021-12-15
Replace Spectral with Optic CI

cmars:
The moment we've all been waiting for! 🎉 This replaces our Spectral-based implementation of our API standards with Optic CI.


jcsackett (commented):
This looks good, but I wonder about the various yaml definitions here in components. Is this the place those are all going to live going forward, because the same definitions live in several repos and drift seems likely, unless drift is flagged as an error if things don't match these components. That may already be in this diff, but it's quite big and if so I missed it. 😝

cmars:
> This looks good, but I wonder about the various yaml definitions here in components. Is this the place those are all going to live going forward, because the same definitions live in several repos and drift seems likely, unless drift is flagged as an error if things don't match these components. That may already be in this diff, but it's quite big and if so I missed it. 😝

For now... might as well define the common API components (JSON API, versioning, other bits) close to the place where we define these standards, but yes, it needs improvement.

Drift is currently a problem and it will only get worse; we're in dire need of a central schema repository of some sort for these -- even something as simple as a static files HTTP site. With versioning. Something I'll be pitching for.

cmars:
It's an interesting idea; if we're compiling and collating and localizing OpenAPI specs all over the place, we're going to need a way to compare the components as you describe. There is some schema checking in these rules (are requests and responses formatted to our flavor of JSON API for example) but not to the degree of catching drift we would need.

It's an interesting alternative -- though it might cause developer pain when you find your components have drifted. Could we prevent it from even happening?

cmars:
JFDI

PR #128, opened 2021-12-15
chore: update CircleCI release configuration

cmars:
This updates CircleCI automated releases to:
- Publish to NPM only on pushed tags
- Publish to Docker on pushed tags and landing in main
- Update the Docker latest tag when the image tag is the latest git
  tag (new tag release).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,8 @@ defaults: &defaults
   docker:
     - image: circleci/node:14-buster
   working_directory: ~/sweater-comb
+  environment:
+    IMAGE: ghcr.io/snyk/sweater-comb
 
 jobs:
   test:
@@ -26,7 +28,7 @@ jobs:
           name: E2E Test
           command: ./end-end-tests/api-standards/test-bulk.bash
 
-  docker-main-release:
+  docker-release:
     <<: *defaults
     steps:
       - checkout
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -26,22 +28,35 @@ jobs:
...
+            [ -n "$TAG" ] || exit 1
+            ./scripts/build-docker.bash'
       - run:
           name: Push Docker image
+          environment:
+            GIT_BRANCH: << pipeline.git.branch >>
+            GIT_TAG: << pipeline.git.tag >>
           command: |-
+            [ -n "$IMAGE" ] || exit 1
+            export TAG=${GIT_TAG:-${GIT_BRANCH}}
```
genslein, line 54: If `GIT_TAG` is not set, I don't think this works. Did some quick bash shelling locally. may just want to use git tag?

genslein (approved):

.circleci/config.yml
cmars: I _think_ it'll be OK. The ${_var_:-_default_} should evaluate to _default_ if _var_ is unset.

```
sh-3.2$ a=foo
sh-3.2$ b=bar
sh-3.2$ echo ${a:-${b:-nope}}
foo
sh-3.2$ echo ${c:-${d:-nope}}  # $c and $d unset
nope
sh-3.2$ echo ${c:-${b:-nope}}
bar
```

PR #132, opened 2021-12-17
fix: use semantic-release with yarn

cmars:
sweater-comb is a yarn project, it needs to be semantic-released with
yarn.


cmars:
JFDI

PR #133, opened 2021-12-17
demo: show team how they will can support multiple rulesets

acunniffe:
Minor version `0.16` includes support for running multiple rulesets -- selected via the `--ruleset` flag

We anticipate adding helpers to turn on/off certain rules, to merge entire rulesets to make a large bank of rules easier to compose into all the rulesets you need across the org. 

- Nothing changes in how vervet calls the rules. By default `--ruleset default`, and the current v3 rules are keyed as `default`. 
- If you were to run compare with `--ruleset internal-apis` all the spectral rules will run, but everyone else has been commented out. 
- This pattern of making a new check service for each ruleset, and giving it a CLI friendly key will scale to as many different sets as you want 

Expecting no tests to change / fail.


src/index.ts
```diff
@@ -1,7 +1,11 @@
...
+import { v3Rules } from "./v3-rules";
+// import { internalApiRules } from './internal-api-rules';
 
-const apiCheckService = newSnykApiCheckService();
-const cli = makeCiCli("sweater-comb", apiCheckService);
+const cli = makeCiCliWithNamedRules("sweater-comb", {
+  default: v3Rules(),
+  "v3-apis": v3Rules(),
+  // "internal-apis": internalApiRules(),
+});
```
acunniffe, line 10: @cmars I commented out `internal-apis` rules so nobody tries using it before it's ready

cmars (approved):

cmars (approved):

cmars:
@acunniffe This looks good. Picking up from our conversation earlier, what we're needing near-term for a separate internal API ruleset, is one based on our public rules, but with some of those rules removed or modified slightly.

For example, we'd want to replace the orgs/groups tenancy rule with one that exempts certain paths which only apply to the internal routes.

CC @gablaxian @ahmed-agabani-snyk

cmars:
Closing this -- this was much appreciated, and we've got awesome multi-ruleset support with the new DSL.

Issue #137, opened 2021-12-17
Missing operation summary fails CI; fixing it triggers "invalid operation ID" (catch-22)


cmars:
When introducing Optic CI into an existing project, in addition to #136 I found these failures:

```
  FAIL   This Specification
  x this specification:  must have an org or group tenant
    expected support for org or group tenant: expected [] to have a length above 0 but got 0
    Read more in our API guide (​https://github.com/snyk/sweater-comb/blob/main/docs/standards.md#organization-and-group-tenants-for-resources​)
    at (/Users/cmars/Projects/registry/src/web/routes/api/v3/resources/_examples/hello_world/2021-06-13/spec.yaml:1:0)


  FAIL  GET /examples/hello_world/{id}
  x requirement for operation: GET /examples/hello_world/{id} must have a summary
    expected undefined to exist
    Read more in our API guide (​https://github.com/snyk/sweater-comb/blob/main/docs/standards.md#operation-summary​)
    at (/Users/cmars/Projects/registry/src/web/routes/api/v3/resources/_examples/hello_world/2021-06-13/spec.yaml:65:2767)


  FAIL  POST /examples/hello_world
  x requirement for operation: POST /examples/hello_world must have a summary
    expected undefined to exist
    Read more in our API guide (​https://github.com/snyk/sweater-comb/blob/main/docs/standards.md#operation-summary​)
    at (/Users/cmars/Projects/registry/src/web/routes/api/v3/resources/_examples/hello_world/2021-06-13/spec.yaml:14:250)
```

Oh, easy fix, I thought, I'll go and add summaries to those operations, it's a non-breaking change, and we'll be good!

Well, no, when I added the summaries, it triggered this rule failure because the operation IDs are also non-standard:

```
  FAIL  GET /examples/hello_world/{id}
  x updated operation: GET /examples/hello_world/{id} must have the correct operationId format
    operationId "helloWorldGetOne" must be camelCase (helloWorldGetOne) and start with get|create|list|update|delete: expected false to be truthy
    Read more in our API guide (​https://github.com/snyk/sweater-comb/blob/main/docs/standards.md#operation-ids​)
    at (/Users/cmars/Projects/registry/src/web/routes/api/v3/resources/_examples/hello_world/2021-06-13/spec.yaml:66:2834)


  FAIL  POST /examples/hello_world
  x updated operation: POST /examples/hello_world must have the correct operationId format
[...]

PR #139, opened 2022-01-04
tune operation ids to prevent catch-22

acunniffe:
Addresses https://github.com/snyk/sweater-comb/issues/137  by tuning rules for operationIds

We should discuss the proposal in the comment -- there's some important choices to make about how we'll handle these sort of things going forwards. `effectiveOn` feels right


acunniffe:
Closed. let's fix this with `effectiveOn`

Issue #136, opened 2021-12-17
Cannot pass CI with pre-existing API spec that lacks tenancy params 


cmars:
When introducing Optic CI into an existing project, I found this rule failing with no changes:

```
  FAIL   This Specification
  x this specification:  must have an org or group tenant
    expected support for org or group tenant: expected [] to have a length above 0 but got 0
    Read more in our API guide (​https://github.com/snyk/sweater-comb/blob/main/docs/standards.md#organization-and-group-tenants-for-resources​)
    at (/Users/cmars/Projects/registry/src/web/routes/api/v3/resources/_examples/hello_world/2021-06-13/spec.yaml:1:0)
```

The paths are: `/examples/hello_world`, `/examples/hello_world/{id}`. They aren't in compliance, but they're also not changed in the CI diff.

We want to improve these, but if they've already landed, they shouldn't block introduction of Optic CI into the project. Tenancy rules may also need to evolve over time...

acunniffe:
Hola! Back from break, hope yours was good. 

Since we're talking about checking for things that need to be present, we have to use `requirement`. We could do more tuning to prevent this from being triggered in this instance, but I'm feeling a strong pull towards the `effectiveOn` idea you shared in December that allows old versions to be exempt from some rules. 

> Tenancy rules may also need to evolve over time...

your callout here reinforces this direction from my POV. 

Should I move ahead and have the team open some PRs for this?

cmars:
Welcome back! It was a good break, hope yours was as nice and relaxing!

Yes please re: `effectiveOn`; that should solve this particular problem nicely, and lets us confidently evolve our rules moving forward.

cmars:
Fixed by #140

PR #140, opened 2022-01-04
allow certain rules to qualify/disqualify themselves based on effectiveOn

acunniffe:
#137 and #136 are both best addressed by exempting older date versions from following all the rules. Any other approach requires us to encode logic into the rules themselves making them harder to write, read and maintain. 

@cmars had an awesome idea weeks ago, that I'm very thankful he thought of: just set certain rules to only take effect after a certain date. 

Snyk is a little different than other users in that the date you probably want to compare against is the date version from vervet, not the date the `optic-ci` check is being executed. Implied in that, is a way to version the rules themselves and say these checks ONLY should apply to date versions that are created after X, exempting ones that came before from having to pass. 

By default each check always applies to all versions, it is only when you constrain it using the metadata argument (same thing you use to link up the docs) that it may become filtered.
```
metadata.docs('https://yourdocs')
metadata.becomesEffectiveOn(new Date('2021-1-10'))
```

Next steps: 
- landing this PR will make this possible w/o any regressions or changes in optic-ci's behavior 
- provide Aidan guidance on which rules should have an effective date and he can add them to this PR (fixing 137 + 136)
- OR someone at Snyk can go add effective dates themselves to this or a future PR

Questions: 
- will we need an end constraint too? Now? later?`stopsBeingEffectiveOn`?
- Should the date we compare against be the date from vervet or the current (what the clock says) date? I think I understood @cmars intent, but want to make sure. It's easy to change


src/service.ts
```diff
@@ -3,7 +3,11 @@ import { ApiCheckService, DslConstructorInput } from "@useoptic/api-checks";
 import { oas } from "@stoplight/spectral-rulesets";
 
 export function newSnykApiCheckService() {
-  const snykRulesService = new ApiCheckService<SynkApiCheckContext>();
+  const snykRulesService = new ApiCheckService<SynkApiCheckContext>(
+    (context) => {
+      return new Date(context.changeDate);
```
acunniffe, line 8: @cmars this is where that key assumption lives re: what date should be used during the comparison. 

If it's the resource date from vervet, this code is right

if it should be today's date, we would replace with `new Date()`

I'm pretty sure you want it as is, but lmk if I misunderstood

cmars: Looks correct as it is, `changeDate` is the date the change should be considered to take place. It could be (and usually is) "right now" but it is useful to be able to control it so that you can set a policy around it, or check how a rule might evaluate in a given scenario. Tests for sure, but also perhaps to answer questions like "I want to do ${crazy_thing_to_the_api} next week, will it be allowed"?

cmars:
> Snyk is a little different than other users in that the date you probably want to compare against is the date version from vervet, not the date the optic-ci check is being executed. Implied in that, is a way to version the rules themselves and say these checks ONLY should apply to date versions that are created after X, exempting ones that came before from having to pass.

If someone wanted a different version scheme, they could probably map that onto dates quite easily. For example, if they used semantic-release, they could derive dates from a map, or directly from git release tags.

cmars (approved):

acunniffe:
Definitely -- I think this is fairly versatile and gives us space to figure out how other people would want to use it, while at the same time, supporting your current need. 

Do you want me to tune the metadata rules with a date? If so what date?

cmars:
> will we need an end constraint too? Now? later? stopsBeingEffectiveOn?

We may want to deprecate a rule like this... if we needed to deprecate a rule so that stops applying to new versions, but continues to apply to existing ones. One example might be a "breaking changes" type of rule -- we don't want to remove a rule that protects against breaking changes in existing versions, but it might no longer apply to newer versions as standards evolve.

> Should the date we compare against be the date from vervet or the current (what the clock says) date? I think I understood @cmars intent, but want to make sure. It's easy to change

I'll try to illustrate the desired behavior here with an example. Given a rule `effectiveOn: '2021-09-20'` and a `context.json`:

```
{
  "changeDate": "2021-11-11",
  "changeResource": "thing",
  "changeVersion": {
    "date": "2021-11-10",
    "stability": "experimental"
  }, ...
```

The rule should apply, because the `rule.effectiveOn && changeVersion.date > rule.effectiveOn`.

Given a context for an older resource version spec,

```
{
  "changeDate": "2021-11-11",
  "changeResource": "older_thing",
  "changeVersion": {
    "date": "2021-06-12",
    "stability": "beta"
  }, ...
```

The rule should not apply, because `changeVersion.date` is before the rule's effectiveOn date.

acunniffe:
Makes sense. I'm hearing 

1. defer the ability to use dates to replicate 
2. Aidan and casey are on the same page about how the dates are compared. This PR will make it work that way.

PR #141, opened 2022-01-05
chore: use effectiveOn for select rules

cmars:
Make the tenancy and operationId naming rules effective since
2021-07-01. This allows some already-released early v3
implementations to pass CI.

This also changes the frame of reference to the version date of the
resource, rather than the date on which the change is occurring. This
allows CI to pass on non-breaking changes to releases made prior to
later rules.

Side note: This does raise the need for a rule which prevents
back-dating a new release past a certain point. There are several
reasons back-dating releases is a problem (besides just circumventing
rules) -- it can play havoc with deprecation & sunsetting, as well as
surprise users. Will open a separate issue!


src/rulesets/operations.ts
```diff
@@ -35,6 +35,7 @@ export const rules = {
       "have the correct operationId format",
       (operation, context, docs) => {
         docs.includeDocsLink(links.standards.operationIds);
+        docs.becomesEffectiveOn(new Date("2021-07-01"));
```
jcsackett, line 38: Why this date instead of, say, today's date? Shouldn't anything added that requires passing this rule have been added starting from today?

I assume not actually, and that I'm missing some subtlety to how the rules around this work.

jcsackett (commented):
LGTM, though I'm confused about the chosen date for this to be effective.

src/service.ts
```diff
@@ -5,7 +5,7 @@ import { oas } from "@stoplight/spectral-rulesets";
 export function newSnykApiCheckService() {
   const snykRulesService = new ApiCheckService<SynkApiCheckContext>(
     (context) => {
-      return new Date(context.changeDate);
+      return new Date(context.changeVersion.date);
```
acunniffe, line 8: 🤔 this makes more sense now. Over the break my brain must have forgotten which part of the context should be used for this.

acunniffe (approved):

acunniffe:
> Side note: This does raise the need for a rule which prevents
back-dating a new release past a certain point. There are several
reasons back-dating releases is a problem (besides just circumventing
rules) -- it can play havoc with deprecation & sunsetting, as well as
surprise users. Will open a separate issue!

This is interesting. It's definitely reasonable to write an `optic-ci` rule that prevents backdating, but I could see it getting kinda dicey since a PR might sit open for days and then end up being implicitly back-dated. Might need context from when the PR was first opened to do this in a reasonable way. Or some policy that all new date versions must be > today's date -- encouraging folks to add some buffer to their new resources. Just my 2 cents.

src/rulesets/operations.ts
cmars: The only resource in our API that fails this rule is our v3 Examples resource, https://apidocs.snyk.io/?version=2021-06-13%7Eexperimental#tag--Examples.

I'm being more conservative here in only exempting releases prior to 2021-07-01 from this rule. If I chose today, this might allow someone to add an operationId with a format we don't want to any active release; adding an operation is a non-breaking change.

We really should deprecate and sunset the above Examples resource from our public API.

cmars:
@acunniffe Having similar thoughts wrt backdating.. such a rule would need to deal with the reality of PRs... and git's interesting construction of timelines. Need to think on this one some more.

src/rulesets/operations.ts
jcsackett: Thanks for the clarification!

PR #143, opened 2022-01-10
Change rules should be more specific about when they should run

acunniffe:
In response to https://github.com/snyk/sweater-comb/issues/137

Metadata rules about things like `operationId` and `summary` are getting triggered on change when their host entity (in this case the operation itself) are changed. For instance, changing `summary`, will trigger the rule that runs when `operationId` is changed today. 

Here we add a new qualifier `.attributes(...)` to `changed` and `requirementOnChange`

Now you can further control when changed rules are triggered. If it's a rule about a particular attribute of an entity like `operationId` for operations or `format` for properties -- it can make sense (although is not required) to constrain the rule further. 

In this case, the rule will be triggered for any operation that changes or is added. If it has changed, it will only consider changes to `operationId`. If another attribute like `summary` changed in the operation the rule is skipped. 

```
operations.requirementOnChange
      .attributes("operationId")
      .must(
```


src/rulesets/headers.ts
```diff
@@ -5,13 +5,12 @@ import { links } from "../docs";
 
 export const rules = {
   headerNameCase: ({ responses }: SnykApiCheckDsl) => {
-    responses.headers.requirementOnChange.must(
-      "be kebab-case",
-      ({ name }, context, docs) => {
+    responses.headers.requirementOnChange
+      .attributes("name")
```
acunniffe, line 9: constrain the header naming rule to changes to its name

**Currently** if you changed the description of a header with an invalid name, the rule would not pass. It's a breaking change to fix a name leading to a catch 22. 

**Now** all the sibling attributes of `name` can change without triggering this rule

src/rulesets/operations.ts
```diff
@@ -31,22 +31,24 @@ const preventParameterChange = (schemaProperty: string) => {
...
-            normalized === operation.operationId &&
-              prefixRegex.test(operation.operationId),
-            `operationId "${operation.operationId}" must be camelCase (${normalized}) and start with get|create|list|update|delete`,
-          ).to.be.ok;
-        }
-      },
-    );
+    operations.requirementOnChange
+      .attributes("operationId")
+      .must(
```
acunniffe, line 36: constrain the operation id naming rule to changes to its operationId

**Currently** if you changed changed anything else in the operation it would run this rule 

**Now**  this only runs if the operationId actually changed

src/rulesets/operations.ts
```diff
@@ -74,18 +76,17 @@ export const rules = {
     );
   },
   removingOperationId: ({ operations }: SnykApiCheckDsl) => {
-    operations.changed.must(
-      "have consistent operation IDs",
-      (current, next) => {
+    operations.changed
+      .attributes("operationId")
+      .must("have consistent operation IDs", (current, next) => {
```
acunniffe, line 81: constrain the operation id naming rule to changes to its operationId

**Currently** if you changed changed anything else in the operation it would run this rule 

**Now**  this only runs if the operationId actually changed

src/dsl.ts
```diff
@@ -177,19 +177,6 @@ export class SnykApiCheckDsl implements ApiCheckDsl {
           ),
         );
       },
-      should: (statement, handler) => {
```
cmars, line 180: I'd like to understand this removal here and below better. Does ShouldOrMust now have a default so that this is no longer necessary?

cmars (approved):
Looks good! Just a (non-blocking) question below.

cmars:
Looks good to land once the conflict is resolved. Thanks for this!

src/dsl.ts
acunniffe: Yes -- `should` never did anything different than `must` so began to deprecate it -- and I updated your DSL to align with that. 

I can see a use for warnings and errors being different, but if we want to bring it back let's do it with intention and have different behaviors planned

PR #148, opened 2022-01-19
Improve failing messages

smizell:
This is a small PR to improve some of the error messages that are hard to understand. The `expect(...).to.not.be.true` was especially hard to follow, so we've changed them to `expect.fail` and added better messages. You can see the difference in messages in the snapshot tests.


cmars (approved):
Looks good, thanks for this!

cmars:
Reopened to try and nudge CircleCI... not sure why it wasn't running.

Issue #147, opened 2022-01-19
PUT not allowed


cmars:
JSONAPI does not allow PUT requests, so we shouldn't either.

acunniffe:
This should be as easy as adding an `operation.added` rule

Since operations are uniquely id'd by their path pattern and method, there's no way to *change the method of an operation. 

want one of us to go add one?

cmars:
That'd be great, thank you!

PR #151, opened 2022-01-23
do not allow adding put operations

acunniffe:
Resolves: https://github.com/snyk/sweater-comb/issues/147

The rule applies on `added` for a few reasons:

- Operations are uniquely identified by combining their `method` and a normalized `path` (ie `/users/{userId}` -> `/users/{}`
- Because the method is part of the unique ID, you can't really 'change' the method. Doing so just shows up as removing the first operation and adding another. Removing would trigger a failure, and adding would trigger the new rule

***for the curious:*** We could make it so that Operation IDs are the unique ID, but then rules about them changing can not be enforced. There's some interesting research in multi factor IDs in AST editors, but we don't need that for our use cases here. Everything works, just wanted to explain why it's an added vs a changed rule.


cmars (approved):
Thanks for this!

Issue #146, opened 2022-01-19
Resources must be plural nouns


cmars:
This will likely require some slight smarts because natural language, but all resource names should be plural nouns.

acunniffe:
Looks like https://www.npmjs.com/package/pluralize might be decent and extensible. The domain has a limited vocabulary so if you had to add a special case, this library supports it. 

- would check this on `operation.added`
- would take the resourceName, `pluralize(resourceName)` it and check if the values are the same

want one of us to take a stab?

cmars:
Yes please! 🙏

PR #152, opened 2022-01-23
Required plural resource names 

acunniffe:
Addresses: https://github.com/snyk/sweater-comb/issues/146

Now all `changeResource` names must be plural. 

***Note:***  It's hard to do a requirement for nouns without a dictionary of some kind, but you could define an allowed list of names in the Snyk domain ie 'issues', 'runs', etc. I chose the example of 'runs' because it's a verb and a noun depending on how it is used. If it's really important to constraint the resource names people choose I strongly suggest a list of allowed terms being added to sweater comb 

The snapshots are messy because we were using resource `Example` everywhere and had to update it to `Examples` to make the rule pass.


src/rulesets/tests/__snapshots__/lifecycle.test.ts.snap
```diff
@@ -827,6 +827,73 @@ Object {
...
+          },
+          "conceptualPath": Array [],
+          "jsonPath": "/",
+          "kind": "ContextRule",
+        },
+      },
+      "condition": "resource names must be plural",
+      "docsLink": undefined,
+      "effectiveOnDate": undefined,
+      "error": "\\"Dog\\" is not plural. Try using \\"Dogs\\"",
```
acunniffe, line 887: ^ Look at that 🥇

src/rulesets/resource-names.ts
```diff
@@ -0,0 +1,17 @@
+const pluralize = require("pluralize");
+
+// Here's your escape hatch to customize plural rules for your domain
+// https://www.npmjs.com/package/pluralize#usage
+
+// pluralize.addUncountableRule("snyk"); // no way to make this plural
+// pluralize.addPluralRule("bacterium", "bacteria");
```
acunniffe, line 7: here is where we can teach it about weird words that may exist in the Snyk domain that don't fit nicely

cmars (commented):
Looks amazing. Want to give this another pass tmw before landing, just to be sure of the impact.

PR #153, opened 2022-01-24
docs: adds policy for settings

gablaxian:
Long have we debated the preferred way to 'handle' settings within our API.
Today, that changes. Rather than use some settings 'area' (`/settings/[resource]`) or some kind of settings relationship on a resource (`/projects/settings`) we are deciding to standardise around settings being an attribute directly on a resource:

```
GET /orgs/{orgId}
{
    "jsonapi": {
        "version": "1.0"
  },
    "data": {
        "type": "org",
        "id": "111-222-333",
        "attributes": {
            "name": "Amazing Org",
            "slug": "amazing-org",
            "settings": {
                "requestAccess": {
                    "enabled": true
                }
            }
        }
    }
}
```


jbowes:
Aesthetically I don't like it, but with that aside, I see a few problems:

gablaxian:
@jbowes Thanks for taking a look and starting the conversation!

I don't think any solution will be without its issues. So let's find the one with the least...

I recall the sub-resource approach being sub-optimal for several reasons:
- It looks like a resource when it (usually) isn't
- It floods our API with many `/settings` endpoints which just seemed not very user friendly.
- It likely still has similar same authz implications for when not all settings are available to all users.
- It doesn't easily allow for creating settings when creating the resource, which is what customers are requesting currently.

docs/standards.md
```diff
@@ -14,6 +14,32 @@ Resources addressed by Group begin with `/v3/groups/{group_id}/...`.
 
 Additional resource properties that must be used in resource attributes, where applicable.
 
+## Handling resource settings
+
+Resource settings are an attribute of the resource and not a resource or relationship. Using Organizations as an example, we would see a response like:
```
ianlivingstone, line 19: I think we need to provide some guidance on how to organize settings - some will be core platform settings while others will be specific to products.

Whatever name spaces we use for product settings should be consist throughout the API and they should be standardized upon in our API Guidelines.

ianlivingstone:
> It doesn't easily allow for creating settings when creating the resource, which is what customers are requesting currently.

This point here is the key behind much of work: enabling customers to programmatically manage Snyk, especially, enterprise customers.  It'd be great to add justification like this to be the above proposal as it's guiding principle for the API proposal.

docs/standards.md
cmars: If we're introducing settings as a core API principle, it needs a really clear and concise definition. Settings are starting to feel like a sort of "structured metadata" about a resource -- a "resource about a resource" or "meta-resource". That may be just fine; this pattern might lead us to the right set of principles defining them.

In considering where settings fit, some questions I'd pose to try and flesh this out:

- What are settings? (as a concept)
- How are they different from regular resources? Or are they themselves resources? What makes settings special?
- What is the subject of a settings instance? Does it describe a resource directly? Or is it more subtle (how a resource should be interpreted or used)?
- How do settings relate to a resource's containment hierarchy?
- What are examples of valid settings targets?

Some arguments against settings, which should be addressed if we decide to adopt them:
- What are some anti-examples of settings? IOW when should settings be avoided, even if they seem like they could solve a problem?
- Why can't settings be a common property in many resources' attributes?
- How do some other RESTish APIs deal with this sort of thing? Compare & contrast.

garyposter: Like Casey, I’m not arguing for a particular perspective. Here are a few different considerations.

First, Casey’s parallel to metadata resonates with me. I can think of two sorts of other metadata-ish stuff that we have in mind.

One is “count” and other information about projects etc. See here: https://www.notion.so/snyk/API-v1-vNEXT-survey-notes-d19388ddd3894ee09268cf8f75d04a81#e9c2f2d7304d45ab8d5eb69b63704144

Another is our desired future annotations. We’ve imagined being able to have annotations on any resource, and in different contexts. So, on a given resource, I might have a personal set of annotations, a GitHub app might have another set of annotations, a runtime app might have a third set of annotations, and a group administrator might have set a fourth set of annotations available to anyone in that group.

Maybe we want annotations, count (and other pre-defined read-only metadata), and settings (yes, as Casey asked, what exactly are these? What are their definitions? Why are they not just attributes of the resource?) to all follow a similar pattern. Maybe?

(I don’t think any of this stuff is equivalent to JSON:API `meta` stuff. That has no defined approach for edits.)

——

Second, I initially imagined that these all would be separate resources. Dunno. 

I looked for some examples. Everything I found was in a separate object. However, I didn’t find much, and nothing specific to JSON:API.

https://developer.wordpress.org/rest-api/reference/settings/
https://developers.arcgis.com/rest/users-groups-and-items/organization-settings.htm
https://devcenter.heroku.com/articles/platform-api-reference#app-feature-update

And hey, here’s how Snyk’s v1 did it! https://www.notion.so/snyk/API-v1-vNEXT-survey-notes-d19388ddd3894ee09268cf8f75d04a81#8841cf013ab247c88692baab0176172f

——

And finally, if we follow the pattern Graham suggests, IMO we should plan to start including JSON API sparse fieldset capabilities. I was hoping to avoid that but Gareth V was pushing for it elsewhere and we may really want it to support GraphQL. 🤷 

And also, in this proposal, how do we send a PATCH to change only one field? Could we see an example of that? I think that’s an important use case from the v1 API survey. See the Snyk v1 settings survey link above. Here’s an (IMO) important excerpt from that: “I think we could standardize a lot of the API with a settings-with-PATCH pattern, and eliminate one-off RPC APIs like "activate" and so on. We could also potentially have a single abstract Node implementation of a settings resource, which would be very, very nice.” Maybe you disagree? But then how would you deal with that old v1 “activate” endpoint if not as an attribute that can be patched?

Thank you for starting this conversation with a proposal, Graham. ❤️

jcsackett:
This is almost a year old without further progress. @jbowes or @gablaxian -- any feelings on where we are with this? I think we've broadly gotten on without a formal settings standard here, but given the state of things with API work I'd like to either formally close this or else get it moved along and into the standards.

PR #156, opened 2022-01-31
Fix bug in tenant verification, and allow untenanted `/orgs` and `/groups`

DarrellMozingo:
See each commit. Also, tests 🎉


mrded (approved):
Looking good to me. I'm surprised we didn't have a test for `specification.ts` already.

gablaxian (approved):

PR #158, opened 2022-02-03
fix: tenancy requirement should apply to additions only

cmars:
This change moves the orgs and groups tenancy rule to operations added,
ensuring new operations will conform, without breaking CI on existing
endpoints which may need to be deprecated and replaced.


cmars:
CC @acunniffe This is the thing I was asking about earlier.

acunniffe:
This looks correct to me

gablaxian (approved):

src/rulesets/operations.ts
```diff
@@ -155,6 +155,25 @@ export const rules = {
...
+        const url = operation.pathPattern;
+        if (
+          !(
+            url.startsWith("/orgs/{org_id}") ||
+            url.startsWith("/groups/{group_id}") ||
+            url === "/orgs" ||
+            url === "/groups"
+          )
+        ) {
+          expect.fail(`expected support for org or group tenant`);
```
DarrellMozingo, line 172: Could we keep the failed endpoints in the error message, for the UX?

src/rulesets/operations.ts
```diff
@@ -155,6 +155,25 @@ export const rules = {
       },
     );
   },
+  orgOrGroupTenant: ({ operations }: SnykApiCheckDsl) => {
```
DarrellMozingo, line 158: So moving this from specifications -> operations makes it magically work? Nice!

DarrellMozingo (commented):
I deffo don't know enough for a ✅  here, but left comments 😄

src/rulesets/operations.ts
cmars: In order to make this rule operate on a delta rather than make it an invariant that applies to all versions, released or not, I need to make it "hang off of" some element that can change in Optic's model of the OpenAPI spec. You're not going to add a new path that does anything without operations in it (hmm, is that a rule? might need to check on this...) in it, so operations seems a natural place to catch it.The other stuff in specifications seem to be invariants.

Moving it was the easy part -- most of my time on this was spent trying to figure out how to write my Jest tests and get them passing 😅.

src/rulesets/operations.ts
cmars: Good call, updated.

PR #163, opened 2022-02-07
Update optic 0.22.5

niclim:
Updated optic-ci to `0.22.5`

There's a couple of changes here, the ones i want to call out are:
- statusCode is now returned as a string (openAPI has valid statusCodes of `4xx` or `default`) - our traverser didn't handle this previously. This should only affect rule definitions, and I believe all the snapshot tests that are updated should cover this?
- As part of the changelog entry point, there'll be a link to a survey to gather user feedback
- Some changes to the traverser interface - these are mainly internal facing changes, but some of these appear to be used in this repo (we're refining what our public interface needs to be, and once we have something stable we'll move to that)

Let me know if you have any questions!

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,10 +38,10 @@
     "typescript": "^4.4.4"
   },
   "dependencies": {
-    "@useoptic/api-checks": "0.21.3",
-    "@useoptic/json-pointer-helpers": "0.21.3",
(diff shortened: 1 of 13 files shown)


cmars (approved):
Thanks for this!

Issue #155, opened 2022-01-25
GET, POST, PATCH responses must not be empty


cmars:
Operations for these HTTP methods must declare a response with content type `application/vnd.api+json` that is:
- Non-empty
- Conforms to JSON API structure

Currently an operation with no content will pass CI. This should only be allowed for DELETE operations.

This can also happen if content types other than `application/vnd.api+json` is declared.

cmars:
Reminder to myself, to add e2e examples of where this is happening...

cmars:
POST requests must require a requestBody with the proper content type (https://jsonapi.org/format/#crud-creating)

PR #169, opened 2022-02-08
turn on json api rules

acunniffe:
While investigating https://github.com/snyk/sweater-comb/issues/155 we realized that the json api rulesets we wrote together were not being run in the v3 rulesets. I touched those rulesets last and must have forgotten to import them -- very sorry.

I'm asing our team to work on some additional tests and cleaning up the error messages here a bit, but first I wanted to at least get these running again.

Spoke with @cmars out of band about making sure we check if any non-compliant APIs got through the last 2 weeks.


src/service.ts
```diff
@@ -38,6 +38,10 @@ export function newSnykApiCheckService() {
     dslConstructor,
     require("./rulesets/specification").rules,
   );
+  snykRulesService.useDslWithNamedRules(
```
acunniffe, line 41: this was the culprit

acunniffe:
@cmars test fix in flight. Looks like our delete example has a body and shouldn't https://app.circleci.com/pipelines/github/snyk/sweater-comb/379/workflows/a655615d-2ab9-49b3-b8d1-82693da53078/jobs/442

acunniffe:
@cmars tests updated. We'll have another set by EOD

acunniffe:
@cmars ready for review, made with love by @smizell w/ my code review

components/headers/headers.yaml
```diff
@@ -69,3 +69,11 @@ InternalGlooOrgIdHeader:
   schema:
     type: string
     format: uuid
+
+LocationHeader:
```
cmars, line 73: Oh, this is a good catch!

components/common.yaml
```diff
@@ -45,6 +45,7 @@ PaginatedLinks:
     last: { $ref: '#/LinkProperty' }
     prev: { $ref: '#/LinkProperty' }
     next: { $ref: '#/LinkProperty' }
+    self: { $ref: '#/LinkProperty' }
```
cmars, line 48: This as well

cmars (approved):
Thanks again for the turnaround on this!

PR #175, opened 2022-02-10
fix: turn on commitlint

cmars:
Because releases are automated with semantic-release, we need
conventional commits.

We'll ratchet up the --from when enough unconventional commits are behind us :)


cmars:
JFDI

PR #176, opened 2022-02-10
fix: JSON API rules are requirementOnChange

cmars:
Making JSON API rules an invariant requirement gets us into a catch-22
situation (invariant meaning, the rule is applied to all specs as they
are at a point in time, rather than on a changeset) where we have
existing APIs out of compliance, but we can't turn on the rule because
they will never pass.

This changes the JSON API rules to "requirementOnChange", so that the
rule only applies to proposed changes in CI.

Also adding the `schemas` directory to the package build, which was
causing problems in our published NPM and Docker packages.

CC @acunniffe @smizell please confirm `requirementOnChange` is appropriate here. Thanks!


src/rulesets/jsonapi.ts
```diff
@@ -95,7 +95,7 @@ export const rules = {
     );
   },
   responseData: ({ responses }: SnykApiCheckDsl) => {
-    responses.requirement.must(
+    responses.requirementOnChange.must(
```
acunniffe, line 98: use `added`

src/rulesets/jsonapi.ts
```diff
@@ -170,7 +170,7 @@ export const rules = {
     );
   },
   selfLinks: ({ responses }: SnykApiCheckDsl) => {
-    responses.requirement.must(
+    responses.requirementOnChange.must(
```
acunniffe, line 173: as written, any changes to they make to the response body, would require people to add self links for CI to pass. 

Is that what you want? 

If you only want new responses to follow this use `added`

src/rulesets/jsonapi.ts
```diff
@@ -249,7 +249,7 @@ export const rules = {
     );
   },
   compoundDocuments: ({ responses }: SnykApiCheckDsl) => {
-    responses.requirement.must(
+    responses.requirementOnChange.must(
```
acunniffe, line 252: as written, any changes to they make to the response body, would require people to add self links for CI to pass.

Is that what you want?

If you only want new responses to follow this use added

src/rulesets/jsonapi.ts
```diff
@@ -268,7 +268,7 @@ export const rules = {
     );
   },
   schemas: ({ responses }: SnykApiCheckDsl) => {
-    responses.requirement.must(
+    responses.requirementOnChange.must(
```
acunniffe, line 271: as written, any changes to they make to the response body, would require people to add self links for CI to pass.

Is that what you want?

If you only want new responses to follow this use added

src/rulesets/jsonapi.ts
```diff
@@ -21,7 +21,7 @@ function loadSchemaFromFile(filename) {
 
 export const rules = {
   statusCodes: ({ operations }: SnykApiCheckDsl) => {
-    operations.requirement.must(
+    operations.requirementOnChange.must(
```
acunniffe, line 24: this is probably fine. It might ask people to add status codes they don't currently support though when they go change those endpoints. 

If we only want to enforce this on new surface area use `added`

src/rulesets/jsonapi.ts
```diff
@@ -77,7 +77,7 @@ export const rules = {
     );
   },
   contentType: ({ responses }: SnykApiCheckDsl) => {
-    responses.requirement.must(
+    responses.requirementOnChange.must(
```
acunniffe, line 80: this is probably fine. It might ask people to add support for your custom JSON content type if they don't already have it. 

If we only want to enforce this on new surface area use added

acunniffe:
Hey @cmars  -- makes sense. I think we assumed these were working with `requirement` because they had not failed in production. Turns out they just were not running. 

I would suggest we use `added` in a few places instead. Otherwise it's going to start asking people who change older endpoints to start following some more JSON API standards. Since it's in the response body, it's unlikely to cause any breaks, but it might be annoying to be told ex-post-facto to now support JSON API. 

Your call -- you, know the state of things better. 

Since we have new date versions coming out all the time, added works best for many of these IMO.

src/rulesets/jsonapi.ts
cmars: Given that adding self links is a non-breaking change... that actually sounds like a really nice forcing function :)

src/rulesets/jsonapi.ts
```diff
@@ -195,7 +195,7 @@ export const rules = {
     );
   },
   pagination: ({ operations }: SnykApiCheckDsl) => {
-    operations.requirement.must(
+    operations.requirementOnChange.must(
```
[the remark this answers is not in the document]
cmars: Good advice. Tempted to require it, but implementing pagination probably needs a followup PR. I think I agree with `added` here.

cmars:
Thanks for the guidance on this @acunniffe ! I think the guiding question here on added vs requirementOnChage is, "how disruptive would this be to a PR?" It's kind of nice to require drive-by fixes that are fairly innocuous (like self links and missing response codes). Pagination is probably a significant miss -- but also something we should contact the API owner on OOB, because without it, we can easily miss our response time SLOs...

Have to say I'm impressed with the amount of flexibility we have in deciding this kind of policy!

PR #181, opened 2022-02-16
Convert JSON:API tests to use expect.fail

smizell:
This improves the error messages for JSON:API rules by using `expect.fail` and `if` statements. This also uncovered a bug for situations where there was no `application/vnd.api+json` for a response. This is the first of a few improvements for the JSON:API rule messages.


src/rulesets/tests/__snapshots__/end-end.test.ts.snap
```diff
@@ -74545,10 +74545,9 @@ Array [
     "condition": "not allow compound documents",
     "docsLink": "https://github.com/snyk/sweater-comb/blob/main/docs/jsonapi.md#compound-documents",
     "effectiveOnDate": undefined,
-    "error": "Cannot read property 'application/vnd.api+json' of undefined",
     "isMust": true,
     "isShould": false,
-    "passed": false,
+    "passed": true,
```
smizell, line 74550: Here you can see the bug in action.

cmars (approved):
LGTM, thanks for this! Couple things blocking CI, then I'm +1 to land this.

* Commitlint (a `git commit --amend` and prefixing the message header with `fix:` should address it)
* Security vulns in deps. I can waive the check to address these in a followup if it's too disruptive.

src/rulesets/tests/__snapshots__/end-end.test.ts.snap
cmars: Nice catch!

PR #186, opened 2022-02-18
chore: update optic dependency to 0.22.9

niclim:
`0.22.9` includes a change with the integration to the changelog - as part of our internal dogfooding, we've noticed that the github post can be noisy especially when unrelated commits are added, but the openapi spec changes.

Ideally CI doesn't trigger this, but that requires more CI logic to detect changes between commits (something that's not always practical). We've updated the logic to generate fingerprints between diffs, and only generate a new comment if there's a diff in the changes (we want a new comment for visibility to new changes that should be re-reviewed).

I also noticed there was a lot of peerdependency warnings in your JS build, which i've resolved with specifying `"@babel/core": "^7.17.5"` - LMK if you'd rather this in a separate PR

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "format": "prettier --write 'src/**/*.{js,ts}'"
   },
   "devDependencies": {
+    "@babel/core": "^7.17.5",
     "@babel/plugin-transform-runtime": "^7.16.0",
     "@babel/preset-env": "^7.16.0",
     "@babel/preset-typescript": "^7.16.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -24,6 +24,7 @@
     "format": "prettier --write 'src/**/*.{js,ts}'"
   },
   "devDependencies": {
+    "@babel/core": "^7.17.5",
```
cmars, line 27: Is this going to be needed at runtime? We distribute a production npm package of this -- I had to add `@babel/runtime` to get that working a little ways back.

cmars (approved):

PR #187, opened 2022-02-22
Convert to one expect per rule

smizell:
Many of the rules have multiple `expect.fail` calls in them. The problem with this is that, if a rule hits an `expect.fail`, it will stop execution. This can lead to surprising results, especially if the subsequent rules also fail. A user could see an error, fix it, then see a new error.

This PR converts the rules to have only one `expect.fail` per rule. This removes this scenario so that the user sees all of the errors on the first run. This also allows us to have more specific names for the rules, which also improves the experience.


cmars (commented):
First pass this looks amazing. Just a minor thing, can you amend your commit with a `feat: ` prefix in the title? This project uses conventional commits to automatically version releases on landing, which is only triggered by a `feat:` or `fix:` prefix. Thanks!

```
$ commitlint --from=HEAD~1
⧗   input: Convert to one expect per rule
✖   type may not be empty [type-empty]

✖   found 1 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
```

cmars:
I'm still not sure why tests aren't running automatically on PRs here -- seems like I have to close and open PRs to get CircleCI's attention.

cmars (approved):

PR #192, opened 2022-03-02
feat: singleton resources

### LINKED ISSUE #71: Document position on updating singleton resources
When updating singleton resources (there can be only one of its type, like the Highlander), `.data.id` should be optional in the PATCH request body. This is a decision to depart from JSON API, which has this as an unconditional MUST.

cmars:
This adds support for singleton resources, as described in #71.

Properties of singletons:

- Path item has an extension value, `x-snyk-resource-aggregation: singleton`
  - Other possible values might be: `standard` (for lack of a better
    word) and `bulk` (to address the case where we do not make
    single resources addressable.
- Singletons' `.data` schema type is `object`.
- Singletons do not allow /resource/{id}.
- Singletons do not specify an `.id` in the resource. Their type alone is
  their identity. This departs slightly from JSON API and that is
  intended.

Fixes #71

### DIFF
--- a/docs/jsonapi.md
+++ b/docs/jsonapi.md
@@ -160,6 +160,30 @@ The error ID should uniquely identify this occurance of the problem. A server-ge
     - `PATCH /things/:id` to modify one
     - `DELETE /things/:id` to remove one
 
+### Resource Aggregation
+
+#### Singleton
+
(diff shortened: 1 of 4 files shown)


cmars:
Superseded by #206

PR #199, opened 2022-03-08
feat: consolidate common components to improve refs

cmars:
This reorganizes common components into a single components file so that
it may be referenced once in an OpenAPI spec using JSON References.

End-to-end tests are updated to use the updated references.

The component name rule is adjusted to allow this change. It might be
tightened up in a followup to check the names of component leaf-nodes.

Credit for this awesome idea to @acunniffe, originally proposed in #198
as a proof-of-concept.


jcsackett (approved):
LGTM!

PR #200, opened 2022-03-15
fix: allow /test path prefix, exempt from tenancy

cmars:


DarrellMozingo (commented):
Don't forget a test :) https://github.com/snyk/sweater-comb/blob/main/src/rulesets/tests/operations.test.ts#L163

Also, could the commit message give more details on the why, and where it's used?

PR #204, opened 2022-03-16
Workflow commands

smizell:
This PR adds a command for `create-resource` and several commands for adding operations to a spec. Both of these are tested against the existing rulesets to make sure they are correct. You can look at these by checking out the `--help` on the CLI.

The workflow for creating a new resource requires you to be in the parent directory of the `resources` directory. The other commands ask for a path to a spec. There are some areas for improvement in the arguments for these workflows.


acunniffe:
@cmars -- having trouble seeing the Snyk status now that failed

Also can you review one of the generated specs and make sure it plays nice with you code generators?

smizell:
I'm closing this PR for now and will reopen immediately from our `opticdev/sweater-comb` repo instead.

PR #205, opened 2022-03-17
Workflow commands

smizell:
This PR adds a command for create-resource and several commands for adding operations to a spec. Both of these are tested against the existing rulesets to make sure they are correct. You can look at these by checking out the --help on the CLI.

The workflow for creating a new resource requires you to be in the parent directory of the resources directory. The other commands ask for a path to a spec. There are some areas for improvement in the arguments for these workflows.


src/workflows/templates/new-resource-spec.ts
```diff
@@ -0,0 +1,39 @@
...
+  spec["x-snyk-api-stability"] = "wip";
+  return spec;
+}
+
+function baseOpenApiSpec(titleResourceName: string): OpenAPIV3.Document {
+  return {
+    openapi: "3.0.3",
+    info: {
+      title: `${titleResourceName} Resource`,
+      version: "3.0.0",
```
cmars, line 26: We're just using "REST" for the version to indicate the "type" of API it is, since the API is multi-versioned.

```suggestion
      version: "REST",
```

cmars (approved):
LGTM! Some observations below, mostly for awareness -- I'm realizing now I may not have communicated the `/v3` to `/rest` rebrand with y'all.

Issue #196, opened 2022-03-03
Bulk POST


cmars:
We're finding that we need a way to express how resources should be aggregated into collections.

### What we have now

- `POST /resource` creates an individual resource, assigning it an ID in a 201 response.
- `GET /resource` responds with a paginated collection of those resources (get-many).
- `GET /resource/{id}` gets one by ID.
- `PATCH /resource/{id}` updates one by ID.
- `DELETE /resource/{id}` removes one by ID.
- Bulk operations `PATCH /resource` and `DELETE /resource` are possible but currently undefined. I don't think our rules prohibit them.

### Bulk creation: how it's different

- `POST /resource` takes an array for `.data`.
  - It responds `204 No Content` to acknowledge that the collection has been updated with the contents.
  - IDs might be assigned by the client or server. The only guarantee in the response is, the requested resources were created.

Otherwise the other normal resource operations are allowed.

Bulk creation is identified by a POST operation that accepts an array of resources in the top-level `.data` attribute.

Issue #195, opened 2022-03-03
Singleton resources


cmars:
We're finding that we need a way to express how resources should be aggregated into collections.

### What we have now

- `POST /resource` creates an individual resource, assigning it an ID in a 201 response.
- `GET /resource` responds with a paginated collection of those resources (get-many).
- `GET /resource/{id}` gets one by ID.
- `PATCH /resource/{id}` updates one by ID.
- `DELETE /resource/{id}` removes one by ID.
- Bulk operations `PATCH /resource` and `DELETE /resource` are possible but currently undefined. I don't think our rules prohibit them.

### Singletons: how they differ

- POST and DELETE are not allowed -- singletons are neither created nor destroyed.
- `GET /resource` responds with a single resource. It has no `id` in the resource document. This is an intentional break with JSON API. `type` and URI alone are enough to identify it, if it is in fact a singleton.
- `PATCH /resource` updates a single resource.
- There's no `/resource/{id}` for this resource.

A singleton might be identified as such with an extension on the path info, `x-snyk-resource-singleton: true`.

```
paths:
  /singleton:
    x-snyk-resource-singleton: true
    get: { ... }
    patch: { ... }
```

PR #206, opened 2022-03-17
Singleton and Batch support

acunniffe:
This PR addresses https://github.com/snyk/sweater-comb/issues/196 and https://github.com/snyk/sweater-comb/issues/195

Let's talk about how it works: 
- the dsl exposes new structural context so operations, requests and responses can check if they are within an `x-snyk-resource-singleton` path
- the dsl exposes new structural context for `requestDataPropertyIsArray` to support the bulk rule. Since there are no extensions here we need some way for responses to know about the request. 

---
Singleton todos: 
- [x] POST and DELETE are not allowed -- singletons are neither created nor destroyed.
- [x] GET /resource responds with a single resource. It has no id in the resource document. This is an intentional break with JSON API. type and URI alone are enough to identify it, if it is in fact a singleton. **(partial) right now we don't enforce the JSON:API schema, but we should have a singleton schema. optic run will help with this**
- [x] PATCH /resource updates a single resource.
- [ ] There's no /resource/{id} for this resource. **not supported yet** @cmars can there be nested singletons? Trying to figure out the right way to write this

Bulk todos:
- [x] POST /resource takes an array for .data.
- [x] It responds 204 No Content to acknowledge that the collection has been updated with the contents.
- [ ] IDs might be assigned by the client or server. The only guarantee in the response is, the requested resources were created. ** @cmars if/how did you imagine this being a rule? **


acunniffe:
@cmars  -- here's a tested first draft, it should work well, esp when paired with scripts for bulk and singletons. 

Best way to test is to add more end-end examples of bulk and singletons. 

We can merge this with the current capabilities or we can address the open todos and then get it in.

cmars:
Still reviewing, but, I can answer some questions:

> There's no /resource/{id} for this resource. not supported yet @cmars can there be nested singletons? Trying to figure out the right way to write this

Right now, there are no nested singletons, and there is a bit of concern about sprawl if we allow them. That said, we might need to add nuance to this position later with a compelling use case. Can we start with "no nested singletons" now, with the ability to make resource-specific exceptions later if we need to?

> IDs might be assigned by the client or server. The only guarantee in the response is, the requested resources were created. ** @cmars if/how did you imagine this being a rule? **

I don't think this can be a rule, it was more of a statement of what to expect -- something for the standards doc, not the code. Sorry that wasn't clearer!

src/dsl.ts
```diff
@@ -328,14 +337,39 @@ export class SnykApiCheckDsl implements ApiCheckDsl {
...
+
+          const dataProperty = jsonPointerHelpers.tryGet(
+            dsl.nextJsonLike,
+            jsonPointerHelpers.compile(dataPath),
+          );
+
+          const requestDataPropertyIsArray =
+            dataProperty.match && (dataProperty.value as any).type === "array";
+
+          console.log(requestDataPropertyIsArray);
```
cmars, line 369: Remove debugging log message?

cmars (commented):
Thanks so much for this! Looks good, and the pattern of extending context established here will be very helpful to us for future rules. I think the only thing we're missing is explicit test cases for valid and invalid examples of bulk POSTs and singletons.

src/dsl.ts
```diff
@@ -328,14 +337,39 @@ export class SnykApiCheckDsl implements ApiCheckDsl {
...
         OpenAPIV3.ResponsesObject
       >(
         OpenApiKind.Response,
         dsl.changelog,
         dsl.nextFacts,
         (response) => `${response.statusCode}`,
-        (location) => dsl.getContext(location),
+        (location) => {
+          const dataPath = [
+            ...jsonPointerHelpers.decode(location.jsonPath).slice(0, 3),
```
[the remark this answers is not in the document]
acunniffe: Yep -- I think we may add more first class support in a future DSL for rules that run for certain kinds of operations. It seems like all the complexity in our rule handlers is because of this indirection. We find out that 'x changed' and then we try to figure out the context of 'x' and apply a rule.

acunniffe:
I'll add some test cases here Monday 👍

acunniffe:
@cmars -- Please confirm these end-end scenarios for singletons and batch posts. I followed the guides to the best of my understanding, but if there's some surface area we need coverage of that isn't included here, please point me in the right direction. 

Previous limitations (unchecked boxes) still apply.

end-end-tests/api-standards/resources/thing/2021-11-10/000-batch-post.yaml
```diff
@@ -0,0 +1,143 @@
...
+              {
+                $ref: "#/components/x-rest-common/headers/DeprecationHeader",
+              }
+            sunset:
+              { $ref: "#/components/x-rest-common/headers/SunsetHeader" }
+            location:
+              { $ref: "#/components/x-rest-common/headers/LocationHeader" }
+          content:
+            application/vnd.api+json:
+              schema: { $ref: "#/components/schemas/ThingResourceResponse" }
```
cmars, line 53: Content should not be allowed if it's a 204.

end-end-tests/api-standards/resources/thing/2021-11-10/002-fail-batch-post.yaml
```diff
@@ -0,0 +1,137 @@
...
+    post:
+      summary: Create a new thing
+      description: Create a new thing
+      operationId: createThing
+      tags:
+        - Thing
+      parameters:
+        - { $ref: "#/components/x-rest-common/parameters/Version" }
+        - { $ref: '#/components/parameters/OrgId' }
+      requestBody:
```
acunniffe, line 23: @cmars  -- Fixed the test cases for this instance. 

Also realized along the way that none of the e2e tests (even those for regular `post`) have a `requestBody

cmars (approved):

PR #211, opened 2022-03-18
fix: lodash.isequal runtime error

cmars:
Execution of `sweater-comb compare` was failing on a runtime error:

```
internal/modules/cjs/loader.js:905
  throw err;
  ^

Error: Cannot find module 'lodash.isequal'
Require stack:
```

without this change.

Let the record show I am not a fan of bundled dependencies, but it's
necessary here due to the conflicts between ajv 6 and 8, both
dependencies of many common tools.


cmars:
JFDI to fix the package -- right now installation is broken

PR #213, opened 2022-03-19
Made workflow interface a lot nicer for users

acunniffe:
Cleaned up a lot of the initial usability feedback from @cmars that was shared in slack
- Can be run from the same directory as `resources` or any sub dir (just like .git or npm scripts)
- Instead of providing a file to `workflow operation` you just specify `[resource] [version]` where version can be left out to default to `latest`. 
- Paths are lower case now
- You only need to specify the singular and plural forms when creating the resource, they are remembered in your spec using extensions in `info`. 

Other PRs will come monday to enable the example update script and make error messages more readable.

PR #221, opened 2022-03-24
fix: `update` command learning from component schema examples

JaapRood:
Optic's `update` command landed in the cli with #213, but didn't work on examples defined in `components > schemas` of an API spec. This new version of `@useoptic/openapi-cli` brings support for this, enabling the imagined workflow for quickly designing the schemas of new resources by example. It ties in with the resolving spec files by resource names, and is a sub-command of `workflow`.

To update a resource's schema by it's example, run:

```
$ sweater-comb workflow update users
Updated /Users/jaaprood-optic/sites/opticdev/sweater-comb/resources/users/2022-03-24/spec.yaml
✅ Applied 8 patches to 1 file generated from 2 examples
```

~~Since the rest of the workflows are enabled by automatic resolving of resource names (and optionally it's version), we'll be looking at how we can enable that as well, but that's probably done best in a follow up PR; the new version enabling these types of updates came with some other internal changes to code also used by checks, so we'd want to migrate that over sooner rather than later and unblock any eventual patches on that end~~. **Edit**: I snuck in update by resource name in, as that turned out easier than I thought it was going to be. Rather than requiring the Optic's internal update command to allow for different resolving implementations, we could simply wrap the provided updateCommand.


JaapRood:
Not sure why these changes have triggered an `ENOENT 2` 🤔

acunniffe:
@JaapRood all green, we started inferring more of the CI config in the latest CLI version and the the end-end tests were picking up CI=true from Circle even thought the tests were not trying to run sweater-comb in a CI pipeline, just sweater-comb as part of a test that ran in a CI pipeline. 

Upstream fixes in https://github.com/opticdev/monorail/pull/1175

@cmars this should be good to merge.

JaapRood:
I snuck in update by resource name in, as that turned out easier than I thought it was going to be. Rather than requiring the Optic's internal `update` command to allow for different resolving implementations, we could simply wrap the provided `updateCommand`.

Since the way the `update` command fully changes due to working by resource name, there is also less of a reason to keep it at the root of the CLI. In testing, it was pretty obvious that it being part of the `workflow` command will make this easier to work with as a whole.

As a result of this, the following now works:

`sweater-comb workflow update users`

end-end-tests/api-standards/test-bulk.bash
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 set -eu
 HERE=$(cd $(dirname $0); pwd)
+export CI=""
```
cmars, line 4: Is this a new requirement for `compare` / `bulk-compare`? Seems like an unexpected change in behavior.

cmars (commented):
I'm concerned that this new version of the optic dependencies may have introduced a regression that could impact our CI.

We definitely do not want sudden failures if we're running offline -- such as in development.

end-end-tests/api-standards/test-bulk.bash
cmars: Ah, missed @acunniffe 's comment above -- we're likely hitting this _because_ we're running this in Circle, but "offline". I'll test this locally.

acunniffe: >  Is this a new requirement for compare / bulk-compare? Seems like an unexpected change in behavior.

It should not impact running locally, unless there is a local `CI=true` environment variable set. 

It should not affect the CI use case today. 

However, we did fix the regression and have a new version coming out that should have the same behaviors as before. If we wait 20-30 mins for that to release, I can update this to remove the need for these changes to the tests

PR #228, opened 2022-03-28
Fixed content type for create / update types

acunniffe:
From session w/ group 6

@cmars  -- take a look at these to make sure the generators are correct. We did not wrap create / update with `data`

### DIFF
--- a/src/workflows/templates/schemas.ts
+++ b/src/workflows/templates/schemas.ts
@@ -64,18 +64,20 @@ export function buildCreateRequestSchema(
 ): OpenAPIV3.SchemaObject {
   return {
     type: "object",
-    required: ["id", "type"],
+    required: ["data"],
     properties: {
-      id: {
-        type: "string",
-        format: "uuid",
-      },
-      type: {},
-      attributes: {
-        $ref: `#/components/schemas/${titleResourceName}CreateAttributes`,
+      data: {
+        type: "object",
+        required: ["type"],
+        properties: {
+          type: refs.schemas.types,
+          attributes: {
+            $ref: `#/components/schemas/${titleResourceName}CreateAttributes`,
+          },
+        },
+        additionalProperties: false,
       },
     },
-    additionalProperties: false,
   };
 }
 
@@ -84,17 +86,23 @@ export function buildUpdateRequestSchema(
 ): OpenAPIV3.SchemaObject {
   return {
     type: "object",
+    required: ["data"],
     properties: {
-      id: {
-        type: "string",
-        format: "uuid",
-      },
-      type: {},
-      attributes: {
-        $ref: `#/components/schemas/${titleResourceName}UpdateAttributes`,
+      data: {
+        type: "object",
+        properties: {
+          id: {
+            type: "string",
+            format: "uuid",
+          },
+          type: refs.schemas.types,
+          attributes: {
+            $ref: `#/components/schemas/${titleResourceName}UpdateAttributes`,
+          },
+        },
+        additionalProperties: false,
       },
     },
-    additionalProperties: false,
   };
 }
 
--- a/src/workflows/templates/operations/create.ts
+++ b/src/workflows/templates/operations/create.ts
@@ -86,7 +86,7 @@ function buildCreateOperation(
     parameters: commonParameters,
     requestBody: {
       content: {
-        "application/json": {
+        "application/vnd.api+json": {
           schema: createRequestSchema,
         },
       },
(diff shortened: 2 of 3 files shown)


src/workflows/templates/schemas.ts
```diff
@@ -64,18 +64,24 @@ export function buildCreateRequestSchema(
...
-      id: {
-        type: "string",
-        format: "uuid",
-      },
-      type: {},
-      attributes: {
-        $ref: `#/components/schemas/${titleResourceName}CreateAttributes`,
+      data: {
+        type: "object",
+        required: ["id", "type"],
```
cmars, line 71: We generally don't want to allow clients to specify IDs. We might need to allow it in certain circumstances but normally we'll want server-assigned IDs.
```suggestion
        required: ["type"],
```

src/workflows/templates/schemas.ts
```diff
@@ -64,18 +64,24 @@ export function buildCreateRequestSchema(
...
+        properties: {
+          id: {
+            type: "string",
+            format: "uuid",
+          },
```
cmars, line 76:
```suggestion
```

src/workflows/templates/schemas.ts
```diff
@@ -64,18 +64,24 @@ export function buildCreateRequestSchema(
...
+          type: {},
```
cmars, line 77: I believe `type` needs a type.

src/workflows/templates/schemas.ts
```diff
@@ -84,17 +90,23 @@ export function buildUpdateRequestSchema(
...
-      attributes: {
-        $ref: `#/components/schemas/${titleResourceName}UpdateAttributes`,
+      data: {
+        type: "object",
+        properties: {
+          id: {
+            type: "string",
+            format: "uuid",
+          },
+          type: {},
```
cmars, line 102: Here as well

src/workflows/templates/schemas.ts
cmars, line 77:
```suggestion
          type:
            type: string
            pattern: "^[a-z][a-z0-9]*(_[a-z][a-z0-9]*)*$"
            description: Resource type
            example: ${resourceName}
```

src/workflows/templates/schemas.ts
```diff
@@ -70,7 +70,7 @@ export function buildCreateRequestSchema(
         type: "object",
         required: ["type"],
         properties: {
-          type: {},
+          type: refs.schemas.types,
```
acunniffe, line 73: @cmars this is the pattern we've been using, it comes from your common definition

cmars: Ah, sorry, missed that!

cmars: I think we want https://raw.githubusercontent.com/snyk/sweater-comb/common-model-v1/components/common.yaml#/schemas/Types -- a `type: string` with a pattern regex.

The linter seems to be failing on an empty type -- updating this might fix it?

cmars (commented):
I'm a little puzzled at the test failures... see prior comments, I think this might be a missed component type mapping?

For `.data.type`, I think the type we want is one of these from common components v1:

```
schemas:
...
  Types:
    type: string
    pattern: "^[a-z][a-z0-9]*(_[a-z][a-z0-9]*)*$"
    example: resource
    # TODO: enum, generated from all known resources
```

That could be represented in Typescript as a string.

acunniffe:
I'm a bit confused. I just ran the end-end tests and snapshot test locally and they passed. I manually tested `add-create` and got this as the `requestBody`

      requestBody:
        content:
          application/vnd.api+json:
            schema:
              type: object
              required:
                - data
              properties:
                data:
                  type: object
                  required:
                    - type
                  properties:
                    type:
                      $ref: '#/components/x-rest-common/schemas/Types'
                    attributes:
                      $ref: '#/components/schemas/UserCreateAttributes'
                  additionalProperties: false
```

That is the correct`$ref`. 

Is the failing linter empty type commitlist? It looks like it's mad about the fact i approved your code suggestions and those are `merge` not commit list. 

<img width="804" alt="Screen Shot 2022-03-28 at 8 01 40 PM" src="https://user-images.githubusercontent.com/5900338/160506700-c8f43d0f-a485-40d9-a598-0159047ebf70.png">
e

acunniffe:
@cmars I squashed the commits and that made commitlist happy. If I'm missing something, lmk and we can make more changes tomorrow.

cmars (approved):
Looks good, thanks for this!

PR #229, opened 2022-03-29
feat: remove tenancy rules

cmars:
The large number of cases where resources do not fit in the /orgs or
/groups tenancy path hierarchy indicates that this is not a good rule.

Endpoints that depart from this convention will need to define how they
relate to orgs and groups for the purpose of authorization, but what
we've found is, this is not an invariant that can be unconditionally
applied to all resources in Snyk's API.

Attempting to do so, and then making constant exceptions in reaction to
special cases, is only avoiding the reality that there needs to be a
more generally useful and pragmatic way to express authorization and
capability context outside a strict path hierarchy.


ianlivingstone (approved):

gablaxian (approved):

mrded (approved):

mrded:
yeah, org and group tenancy is a big hacky

jcsackett (approved):
I know this will make a lot of people happy. Do you think we can update the workflows to not automatically add the `/orgs/{org_id}/` prefix too? I know in my workshop session that came up around this exact issue.

Or do we want to keep that as the default approach so you do have the little bit of friction of removing that prefix each time, making it clear that you're engaging in an exception?

cmars:
@jcsackett Tenancy will continue to be a thing for a while, until we figure out the authn/authz contextual open-questions. I'm thinking of some kind of option to allow choice here for now, `sweater-comb workflow --tenancy orgs|groups|none` perhaps. Which answers some of the feedback from the workshop as well -- there was uncertainty there whether groups were supported.

PR #230, opened 2022-03-31
chore: bump optic to 0.24.6

niclim:
PR bumps optic from `0.24.0` -> `0.24.6`. 

Downstream changes are:
- Specification facts are now emitted (the metadata on the spec like `info` or `servers`) - this means the changelog will now render metadata changes as well (there's some follow up work to render rule failures on these nodes given how some of the rules around lifecycles are written as well, as well as some rendering improvements we'll be making following this).
- Changed some internal typings for better typescript inferences (`IFact` and `IChange`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,12 +1,9 @@
 {
   "name": "@snyk/sweater-comb",
-  "packageManager": "yarn@3.0.2",
   "version": "1.0.9",
   "main": "build/index.js",
   "types": "build/index.d.ts",
-  "bin": {
-    "sweater-comb": "build/index.js"
-  },
+  "bin": "build/index.js",
   "files": [
     "/build",
     "/schemas"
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -1,12 +1,9 @@
 {
   "name": "@snyk/sweater-comb",
-  "packageManager": "yarn@3.0.2",
```
niclim, line 3: I'm removing this, I'm not sure what was happening before, but the `yarn.lock` file was yarn v1 (so technically this project was always on yarn v1), and for some reason only now my yarn runner wanted to upgrade to v3.

To keep this PR focused, I'm removing this so that the lock file stays on v1 and that other people running this won't upgrade to yarn v3 unless it's intentional

cmars (approved):

PR #232, opened 2022-04-05
feat: automatically discover default vervet resources directory

cmars:
Use vervet project configuration to locate where resources are defined
for workflow commands.

### DIFF
--- a/src/workflows/file-resolvers.ts
+++ b/src/workflows/file-resolvers.ts
@@ -1,18 +1,23 @@
-import findParentDir from "find-parent-dir";
 import fs from "fs-extra";
 import path from "path";
 import { OpenAPIV3 } from "@useoptic/openapi-utilities";
+import { loadVervetResoucePaths } from "./vervet-resolver";
 
 export async function resolveResourcesDirectory(
   workingDirectory: string = getSweaterCombWorkingDirectory(),
 ): Promise<string | undefined> {
-  return new Promise((resolve) => {
-    findParentDir(workingDirectory, "resources", function (err, dir) {
-      if (err || !dir) {
-        resolve(undefined);
-      } else resolve(path.join(dir, "resources"));
-    });
-  });
+  const vervetPaths = await loadVervetResoucePaths(workingDirectory);
+  if (!vervetPaths) {
+    return undefined;
+  }
+  const resolvedWorkingDirectory = path.resolve(workingDirectory);
+  for (const rcPath of vervetPaths.resourcesPaths.values()) {
+    const absRcPath = path.resolve(path.join(vervetPaths.projectRoot, rcPath));
+    if (resolvedWorkingDirectory.startsWith(absRcPath)) {
+      return absRcPath;
+    }
+  }
+  return path.join(vervetPaths.projectRoot, vervetPaths.defaultResourcesPath);
 }
 
 export type ResourceVersionLookupResults =
@@ -53,7 +58,7 @@ export async function resolveResourceVersion(
   const resourceNameLowerCase = resourceName.toLowerCase();
 
   const resourceNames = (await fs.readdir(resources)).filter((maybeResource) =>
-    fs.lstatSync(path.join(resources, maybeResource)).isDirectory(),
+    fs.lstatSync(path.join(resources!, maybeResource)).isDirectory(),
   );
 
   if (!resourceNames.includes(resourceNameLowerCase)) {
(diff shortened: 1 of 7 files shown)


src/workflows/vervet-resolver.ts
```diff
@@ -0,0 +1,41 @@
...
+    } catch (err) {
+      // Do nothing
+    }
+    cwd = path.join(cwd, "..");
+  }
+  return null;
+};
+
+export const getDefaultResourcePath = (): string | null => {
+  const vervetConfPath = resolveContainingPath(".vervet.yaml");
```
cmars, line 21: find-parent-dir doesn't work here, not sure why.

acunniffe:
This looks good so far. I like the approach to testing you took here as well. 

Lmk if you’d like any help with error handling. I think we should probably return / error early if you’re not in a vervet project

cmars:
Updated, PTAL.

src/workflows/tests/vervet-resolver.test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  it("requires a vervet config", async () => {
+    // arrange
+    process.chdir(tmpDir);
+    fs.unlinkSync(".vervet.yaml");
+    // act
+    const vervetPaths = await loadVervetResoucePaths();
+    // assert
+    expect(vervetPaths).toBeFalsy();
+  });
```
carissadean, line 66: Is it possible to add tests for the cases in vervet-resolver when the vervet config is invalid? 
- when the yaml doesn't contain 'apis'
- when the yaml doesn't contain 'resources'

carissadean (approved):
LGTM - added one comment, but not blocking

src/workflows/file-resolvers.ts
```diff
@@ -2,19 +2,23 @@ import findParentDir from "find-parent-dir";
...
-      if (err || !dir) {
-        resolve(undefined);
-      } else resolve(path.join(dir, "resources"));
-    });
-  });
+  const vervetPaths = await loadVervetResoucePaths(workingDirectory);
+  if (!vervetPaths) {
+    return undefined;
+  }
+  workingDirectory = path.resolve(workingDirectory);
```
mrded, line 14: nit: I suggest making a new variable here, and not resetting an argument.

src/workflows/tests/vervet-resolver.test.ts
```diff
@@ -0,0 +1,67 @@
+import * as fs from "fs";
+import * as path from "path";
+import * as os from "os";
+import { loadVervetResoucePaths } from "../vervet-resolver";
+
+describe("vervet init", async () => {
+  const originalwd = process.cwd();
+  var tmpDir: string = "";
```
mrded, line 8:
```suggestion
  let tmpDir: string = "";
```

acunniffe:
LGTM -- Nice work re-implementing the same interface the workflows were using. They should run everyone now once this is merged. I might ask for better error managing but it's not as important since we have this setup to work more seamlessly for everyone. 

Reminder to:
[] update the documentation

src/workflows/file-resolvers.ts
```diff
@@ -2,19 +2,23 @@ import findParentDir from "find-parent-dir";
 import fs from "fs-extra";
 import path from "path";
 import { OpenAPIV3 } from "@useoptic/openapi-utilities";
-import { LogUpdatingSpecification } from "./logs";
-import chalk from "chalk";
+import { loadVervetResoucePaths } from "./vervet-resolver";
 
 export async function resolveResourcesDirectory(
   workingDirectory: string = getSweaterCombWorkingDirectory(),
```
acunniffe, line 8: @cmars do we still need this or do we always want to assume it's in PWD? Will it be set via the env to be the project root somehow?

acunniffe (approved):

src/workflows/tests/vervet-resolver.test.ts
cmars: Good call. TIL... https://stackoverflow.com/questions/762011/whats-the-difference-between-using-let-and-var

src/workflows/file-resolvers.ts
cmars: I think I'll leave it for now. It might not be set, in which case the default is "use the first resources directory in the first API". Setting it allows one to:
* Select a different API / resources directory within the same project
* Operate on a project elsewhere, when your CWD isn't in the project

mrded:
@cmars sorry, this change https://github.com/snyk/sweater-comb/pull/233 with add some extra work to please the linter

mrded (approved):

PR #237, opened 2022-04-08
fix: singletons are not paginated

cmars:
Singletons are not paginated, and pagination parameters should not be
allowed on singletons.

### DIFF
--- a/end-end-tests/api-standards/resources/thing/2021-11-10/002-fail-singleton-no-pagination.yaml
+++ b/end-end-tests/api-standards/resources/thing/2021-11-10/002-fail-singleton-no-pagination.yaml
@@ -1,0 +1,140 @@
+openapi: 3.0.3
+x-snyk-api-stability: experimental
+info:
+  title: v3
+  version: 3.0.0
+servers:
+  - url: https://api.snyk.io/v3
+    description: Public Snyk API
+tags:
+  - name: Thing
+    description: Short description of what Thing represents
+paths:
+  /thing:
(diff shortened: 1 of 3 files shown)


acunniffe:
This makes sense, are there any other places we should be adjusting the rules? I don't want to do them one-off if there are any others we should handle.

cmars:
I don't think so. This one in particular came up when a partner tried to get their singleton API passing lint, and got something like the failure below -- I think it's just a condition we missed in the implementation of singleton. I completely missed this in review as well.

What's weird is, I would have expected the e2e test `001-singleton.yaml` to trigger it.

```
    x added operation: GET /orgs/{org_id}/XXX must correctly support pagination parameters
      expected operation /orgs/{org_id}/XXX get to support pagination parameters, missing: starting_after, ending_before, limit
      Read more in our API Guide (https://github.com/snyk/sweater-comb/blob/main/docs/jsonapi.md#pagination-parameters)
```

acunniffe:
> What's weird is, I would have expected the e2e test 001-singleton.yaml to trigger it.

I think it's because we compare `001-singleton.yaml` to itself. If we had an empty spec, it would be caught since this was an `added` rule. 

Might be worth making that update to the e2e tests.

cmars:
@acunniffe That was it -- making that one a "from scratch" compare triggered it, and a found another rule where I needed to account for singletons (pagination links).

jcsackett (approved):

Issue #238, opened 2022-04-08
Snake case rule should allow numbers in names


cmars:
Feedback from the field (paraphrasing):

> I'm seeing the [snake_case check here](https://github.com/snyk/sweater-comb/blob/main/src/rulesets/properties.ts#L59) doesn't support numbers. We have a couple properties in our API that use numbers (something_in_first_30_days and something_in_past_30_days). Should we change these to spell out the numbers (thirty) or should numbers be allowed in property names?

I think we should allow numbers for word 2..n in a snake case concatenation of words (and in general for any camel/kebab/pascal case).

PR #239, opened 2022-04-11
allow numeric components in snake case

acunniffe:
Fixes https://github.com/snyk/sweater-comb/issues/238

Updated regex should allow names like: `something_in_first_30_days`, but will not allow `30_day_setup`. 

@cmars Is this correct behavior or do you want it to allow digits in all components? 

I defaulted to this approach first because most programming languages do not like variable names that start with digits. It can be annoying for consumers who may discover they need to do some additional mapping while marshalling/unmarshalling to get compilers to work.


src/rulesets/properties.ts
```diff
@@ -56,7 +56,7 @@ export const rules = {
   propertyKey: ({ bodyProperties }: SnykApiCheckDsl) => {
     bodyProperties.added.must("have snake case keys", ({ key }, context) => {
       // TODO: did not find a doc link for this
-      const snakeCase = /^[a-z]+(?:_[a-z]+)*$/g;
+      const snakeCase = /^[a-z]+(?:_[a-z\d]+)*$/g;
```
acunniffe, line 59: small change

acunniffe: Matchers within square brackets are called character classes. A character range is valid inside of a character class. So are the shorthands like `\d`. They're meant to be mixed. source https://www.regular-expressions.info/charclass.html

acunniffe:
@cmars added tests, missed the notif.

cmars (approved):
LGTM & thanks for this! Looks like CI is failing on lint, this just needs a `yarn run format`.

PR #241, opened 2022-04-12
feat: allow breaking changes, immediate sunset of experimental

cmars:
Experimental versions may be sunset without notice.

Breaking changes are allowed in experimental versions without notice.


cmars:
Probably needs some tests, snapshot update.

carissadean (approved):
LGTM - but consider adding tests for the operations.ts and properties.ts files for the cases where the version is experimental and a breaking change is allowed

Issue #256, opened 2022-05-02
Cannot remove experimental API


cmars:
When removing an experimental API (should be allowed now), I'm getting failures due to what appear to be "requirement"-type rules:

```
Comparing (commit e191d9d28cbb59b062d9fbe5137967659d32dde0):internal/rest/resources/repos/2022-04-04/spec.yaml to Empty Spec

   PASS  GET /orgs/{org_id}/repos/{repo_id}


   PASS  DELETE /orgs/{org_id}/repos/{repo_id}


   PASS  GET /orgs/{org_id}/repos


   PASS  POST /orgs/{org_id}/repos


   FAIL  Specification
    x this specification:  must list the available versioned OpenAPI specifications
      expected [] to include '/openapi'
      Read more in our API Guide (https://github.com/snyk/sweater-comb/blob/main/docs/standards.md#making-the-openapi-specification-available)
      at empty.json:1:0

    x this specification:  must provide versioned OpenAPI specifications
      expected [] to include '/openapi/{version}'
      Read more in our API Guide (https://github.com/snyk/sweater-comb/blob/main/docs/standards.md#making-the-openapi-specification-available)
      at empty.json:1:0

    x requirement  must oas3-api-servers
      OpenAPI "servers" must be present and non-empty array.
      at empty.json:1:0



   FAIL  Specification
    x published stability:  must be provided for every resource document
      undefined must be one of allowed values wip,experimental,beta,ga
      Read more in our API Guide (https://github.com/snyk/sweater-comb/blob/main/docs/version.md#stability-levels)

```

While I *could* hack around this in vervet, I don't think we should -- Optic CI should be able to allow the removal of an API spec that satisfies the versioning lifecycle rules.

Issue #255, opened 2022-04-29
Problems with hard-to-read Sweater Comb errors


cmars:
#254 demonstrates two test cases where `compare` fails with a hard-to-read error message. In one case, the OpenAPI is valid but missing a required field (items for an array). In another, the OpenAPI is not valid (wrong type for items).

PR #258, opened 2022-05-03
chore: Bump optic to 0.24.18

niclim:
This PR includes an optic bump + some related fixes:
- Fixes https://github.com/snyk/sweater-comb/issues/256 - requirement rules are not run on removed specifications - I've checked the results / assertions here but please let me know if I've missed anything

- Improved error messages https://github.com/snyk/sweater-comb/issues/255 - examples here https://github.com/snyk/sweater-comb/pull/254 - fixed in `0.24.16` 

Example error response:
```
Error: [
  {
    "message": "Invalid value at path",
    "instancePath": "/paths/~1orgs~1{org_id}~1thing/get/responses/200/content/application~1vnd.api+json/schema/properties/data/items",
    "value": "no"
  },
  {
    "message": "Invalid value at path",
    "instancePath": "/components/schemas/ThingCollectionResponse/properties/data/items",
    "value": "no"
  },
  {
    "message": "Invalid value at path",
    "instancePath": "/components/schemas/ThingCollection/items",
    "value": "no"
  }
]
```


src/dsl.ts
```diff
@@ -254,40 +259,26 @@ export class SnykApiCheckDsl implements ApiCheckDsl {
...
-          ),
-        );
-      },
-    };
+    const dsl = this;
 
-    return {
-      requirement: value,
-    };
+    return genericEntityRuleImpl<
```
niclim, line 264: We now model specifications as a "Fact" which means we can use the generic rule implementation. In the future, we're going to move this into optic internals so the consuming code doesn't need to contain implementation details.

The effect of this change is some _slightly_ different error messages - see snapshot changes for details (`this specification` -> `requirement for specification`)

cmars (approved):
Thanks for this!

PR #260, opened 2022-05-12
chore: bump optic to 0.25.2

niclim:
Bumps optic to 0.25.2 - this includes the new rules authoring experience + new runner.

In this PR I've updated to use the new runner, which is a prerequisite for running the new rules - the functionality should be identical here

I'll be converting rules over slowly side by side, but only switching over the trigger when all the rules are migrated (i.e. the old rules will run until we're ready to switch over).


cmars (approved):
LGTM. Test failures are all due to snapshot changes.

cmars:
There are also some eslint errors (`yarn lint` for details).

niclim:
ok this should be fixed now - snapshot changes were due to us not triggering unnecessary snapshot tests somewhere in our code.

second failure was due to commitlint - both should be fixed now

PR #266, opened 2022-05-13
chore: migrate lifecycle rules to rule authoring v2

niclim:
Migrating the [lifecycle rules](https://github.com/snyk/sweater-comb/blob/main/src/rulesets/api-lifeycle.ts) in this PR

### DIFF
--- a/src/rulesets-new/__tests__/lifecycle-rules.test.ts
+++ b/src/rulesets-new/__tests__/lifecycle-rules.test.ts
@@ -22,7 +22,7 @@ describe("lifecycle rules", () => {
         };
         const results = ruleRunner.runRulesWithFacts(ruleInputs);
 
-        expect(results.length > 0).toBe(true);
+        expect(results.length).toBeGreaterThan(0);
         expect(results.every((result) => result.passed)).toBe(true);
         expect(results).toMatchSnapshot();
       },
@@ -39,7 +39,7 @@ describe("lifecycle rules", () => {
       };
       const results = ruleRunner.runRulesWithFacts(ruleInputs);
 
-      expect(results.length > 0).toBe(true);
+      expect(results.length).toBeGreaterThan(0);
       expect(results.every((result) => result.passed)).toBe(false);
       expect(results).toMatchSnapshot();
     });
@@ -61,7 +61,7 @@ describe("lifecycle rules", () => {
       };
       const results = ruleRunner.runRulesWithFacts(ruleInputs);
 
-      expect(results.length > 0).toBe(true);
+      expect(results.length).toBeGreaterThan(0);
       expect(results.every((result) => result.passed)).toBe(true);
       expect(results).toMatchSnapshot();
     });
@@ -83,7 +83,7 @@ describe("lifecycle rules", () => {
       };
       const results = ruleRunner.runRulesWithFacts(ruleInputs);
 
-      expect(results.length > 0).toBe(true);
+      expect(results.length).toBeGreaterThan(0);
       expect(results.every((result) => result.passed)).toBe(false);
       expect(results).toMatchSnapshot();
     });
@@ -100,7 +100,7 @@ describe("lifecycle rules", () => {
         context,
       };
       const results = ruleRunner.runRulesWithFacts(ruleInputs);
-      expect(results.length > 0).toBe(true);
+      expect(results.length).toBeGreaterThan(0);
       expect(results.every((result) => result.passed)).toBe(false);
       expect(results).toMatchSnapshot();
     });
@@ -137,7 +137,7 @@ describe("lifecycle rules", () => {
           context: sunsetContext,
         };
         const results = ruleRunner.runRulesWithFacts(ruleInputs);
-        expect(results.length > 0).toBe(true);
+        expect(results.length).toBeGreaterThan(0);
         expect(results.every((result) => result.passed)).toBe(false);
         expect(results).toMatchSnapshot();
       });
@@ -152,7 +152,7 @@ describe("lifecycle rules", () => {
           context: { ...sunsetContext, changeDate: "2022-02-01" },
         };
         const results = ruleRunner.runRulesWithFacts(ruleInputs);
-        expect(results.length > 0).toBe(true);
+        expect(results.length).toBeGreaterThan(0);
         expect(results.every((result) => result.passed)).toBe(true);
         expect(results).toMatchSnapshot();
       });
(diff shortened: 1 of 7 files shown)


src/rulesets-new/lifecycle-rules.ts
```diff
@@ -0,0 +1,112 @@
+import { RuleError, Ruleset, SpecificationRule } from "@useoptic/rulesets-base";
+import { links } from "../docs";
+import { stabilityKey } from "./constants";
+
+const stabilityRequirement = new SpecificationRule({
+  name: "resource stability",
+  docsLink: links.versioning.stabilityLevels,
+  matches: (specification, rulesContext) =>
+    rulesContext.specification.change !== "removed",
```
niclim, line 9: matches determines whether a block will run or not

src/rulesets-new/lifecycle-rules.ts
```diff
@@ -0,0 +1,112 @@
...
+  rule: (specificationAssertions) => {
```
niclim, line 10: This is updated that we now model stability on the specification object, rather than exposing a custom written stability assertion

src/rulesets-new/lifecycle-rules.ts
```diff
@@ -0,0 +1,112 @@
...
+      if (diffDays < requiredDays) {
+        throw new RuleError({
+          message: `expected ${stability} resource ${resourceName} to be deprecated ${requiredDays} days`,
+        });
+      }
+    });
+  },
+});
+
+export const lifecycleRuleset = new Ruleset({
```
cmars, line 109: Could I register this with the same kind of call as the old rulesets, `snykRulesService.useDslWithNamedRules`?

src/rulesets-new/lifecycle-rules.ts
```diff
@@ -0,0 +1,112 @@
...
+    );
+  },
+  rule: (specificationAssertions, rulesContext) => {
+    specificationAssertions.requirement("follow sunset rules", () => {
+      const {
+        resourceVersions,
+        changeResource: resourceName,
+        changeVersion: { date, stability },
+        changeDate,
+      } = rulesContext.custom;
```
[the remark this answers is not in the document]
cmars: I see, found it at https://github.com/opticdev/optic/blob/5df2fb2231deadc6f2e3a88d57bbdabb3db61f96/projects/rulesets-base/src/types.ts#L16

This file (types.ts) looks very helpful to understand the shape of objects used in expressing matches and rules. It really seems like the heart and soul of the DSL... could we document these types and their fields with their purpose and intended use? I think the biggest problem I had when trying to write the old rules was, not knowing what the possible parameters to functions were, or what they meant, and trying to infer it from existing examples.

I will say the [README](https://www.npmjs.com/package/@useoptic/rulesets-base) is a huge improvement and such a great start! I'm just thinking of that 5-10% case where we need to dig a little deeper when doing something novel and there's no example to copy from.

src/rulesets-new/lifecycle-rules.ts
cmars: I'm also wondering, how might we select different rulesets from the CLI by name when running compare or bulk-compare? This next cycle, I'm going to need to add a ruleset for linting compiled OpenAPI for gating deployments. Most of the rules can be reused, but a few will be different.

src/rulesets-new/lifecycle-rules.ts
```diff
@@ -0,0 +1,112 @@
...
+});
+
+const stabilityTransitions = new SpecificationRule({
+  name: "resource stability transitions",
+  docsLink: links.versioning.promotingStability,
+  rule: (specificationAssertions) => {
+    specificationAssertions.changed(
+      "not change unless it was wip",
+      (before, after) => {
+        const beforeStability: string | undefined = before.value[stabilityKey];
```
cmars, line 35: What is the shape of `before` here?

How would I know to look in `before.value` for the top-level extension? Looking at the type, it seems like it's an annotated `any` type (unless I read something wrong).

Thinking about this from these perspectives:
- How I would write a specification rule that required the top-level OpenAPI 3 property `tags:` to be a non-empty array?
- How would I make an assertion on the URLs in the `servers:` section?

cmars (approved):
LGTM! Nothing blocking but lots of questions below, trying to really understand what is going on here and how I might use the new DSL. Thanks for this!

src/rulesets-new/__tests__/lifecycle-rules.test.ts
```diff
@@ -0,0 +1,161 @@
...
+        const ruleInputs = {
+          ...TestHelpers.createRuleInputs(baseJson, {
+            ...baseJson,
+            [stabilityKey]: stability,
+          } as OpenAPIV3.Document),
+          context,
+        };
+        const results = ruleRunner.runRulesWithFacts(ruleInputs);
+
+        expect(results.length > 0).toBe(true);
```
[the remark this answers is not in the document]
cmars: It might be potato-potatoe but I've seen other test libraries (in other languages) make semantic use of the comparators when representing failures so I'm curious.

src/rulesets-new/lifecycle-rules.ts
niclim: Few questions:
- How are you invoking different rulesets? There are a few ways I have in mind we could do this (we could set env variables, and parse them to apply different rulesets in `src/index.ts`
- When is this next cycle / when are you planning on starting? Are you going to use these new rule runner? I've got a few more PRs to throw up but I should have this up tuesday afternoon, then we can do some testing + regression tests, etc and make sure this works e2e (just want to know if your wanting to use this earlier)

src/rulesets-new/lifecycle-rules.ts
niclim: before + after are the same shape (i.e. they are both specification facts) - on a changed value, we have both before and after, other assertions only have a single value.

Each value has the following shape (depending on where they are):
```
specificationAssertions."assertion"('condition', (specification)) 

specification = {
  location: - location data for the "fact"
  value: - fact value
  raw: - the raw value at the the fact (i.e. the raw OpenAPI as JSON at that point)
}
```

> How I would write a specification rule that required the top-level OpenAPI 3 property tags: to be a non-empty array?
```javascript
specificationAssertions.requirement(specification => mustBeArray(specification.value.tags))
```

> How would I make an assertion on the URLs in the servers: section?
```javascript
specificationAssertions.requirement(specification => someAssertion(specification.value.servers))
```

src/rulesets-new/lifecycle-rules.ts
cmars: The next cycle starts officially Wednesday. An environment variable would work, if not a command-line switch.

cmars: Another option might be an alternative main entry point?

niclim: > if not a command-line switch

This might be a little harder due to how we initialize and parse the CLI.

Env variable [if statements here](https://github.com/snyk/sweater-comb/blob/main/src/index.ts#L25-L31) would probably be the best option

Something like `process.env.LINT_ENV === 'compiled' ? [...CompiledChecks] : [...NormalChecks]`

PR #268, opened 2022-05-16
chore: migrate header rules to rule authoring v2

niclim:
Migrating the [header rules](https://github.com/snyk/sweater-comb/blob/main/src/rulesets/headers.ts) in this PR


src/rulesets-new/header-rules.ts
```diff
@@ -0,0 +1,44 @@
...
+      if (paramCase(parameter.value.name) !== parameter.value.name) {
+        throw new RuleError({
+          message: `${parameter.value.name} is not kebab-case`,
+        });
+      }
+    };
+    responseAssertions.header.added("be kebab-case", parameterTest);
+    responseAssertions.header.changed("be kebab-case", (before, after) =>
+      parameterTest(after),
+    );
```
niclim, line 19: It's worth noting that we've decided to not include `requirementOnChange` - it's not immediately clear from the wording how this is different, and semantically this means added or changed. As most cases the reason you'd want to have a requirementOnChange is to exempt legacy APIs

With these new rulesets, you can either add your own labels / matchers to explicitly ignore certain operations (`x-legacy: true` or other extentions), or do what I've done here and split this out into added or changed.

cmars (approved):

PR #269, opened 2022-05-16
chore: migrate property rules to rule authoring v2

niclim:
Migrating the [properties rules](https://github.com/snyk/sweater-comb/blob/main/src/rulesets/properties.ts) in this PR

In this case, there's probably better groupings we could have (we could group all breaking changes in a single `Ruleset` and apply a `matches` block for these changes) - we can do this in a follow up PR


src/rulesets-new/property-rules.ts
```diff
@@ -0,0 +1,393 @@
...
+  rule: (responseAssertions) => {
+    responseAssertions.property.added("have snake case keys", (property) => {
+      if (!snakeCase.test(property.value.key)) {
+        throw new RuleError({
+          message: `expected ${property.value.key} to be snake case`,
+        });
+      }
+    });
+  },
+});
```
niclim, line 37: You'll notice that we've split out the request and response property assertions. This is because in a lot of cases, you want to have different rules for request and response properties (e.g. adding a required request body property is a breaking change, but adding a required response body property is not).

There are obviously cases where you want these to have the same rule, but we're leaning towards verbosity, and there are ways we we can share code blocks between these functions.

src/rulesets-new/property-rules.ts
```diff
@@ -0,0 +1,393 @@
...
+}
+
+const enumOrExample = new RequestRule({
+  name: "request property enum or example",
+  docsLink: links.standards.formats,
+  rule: (requestAssertions) => {
+    requestAssertions.property.added("have enum or example", (property) => {
+      const type = property.value.flatSchema.type;
+      if (!withinAttributes(property.location)) return;
+      if (type === "object" || type === "boolean") return;
```
niclim, line 126: The reason we can't add this in the matches block, is because this runs on _every_ property. The nuance is kind of hard to communicate, but on matches, you are matching the top level item (e.g. RequestRule will match a request), where as requestAssertions.properties, are property assertions for the request

cmars (approved):
Looks good!

PR #270, opened 2022-05-16
chore: migrate jsonapi rules to rule authoring v2

niclim:
Migrating the [jsonapi rules](https://github.com/snyk/sweater-comb/blob/main/src/rulesets/jsonapi.ts) in this PR. 

There weren't direct 1-1 tests for these, so I ended up writing a bunch of tests for these. I also split this up into multiple files since there's distinct ideas here (it's not the best grouping but it maintains a 1-1 rule mapping from the old to new rules)


niclim:
Note that I've copied over the rules 1:1 but I think there are more logical groupings that we could have here. For example, we could have either a ruleset for POST requests, or Batch POST requests and then appropriately scope the rules to only trigger when certain conditions match (e.g. requestDataIsArray)

This way, we can colocate singular ideas so it'd be easier to understand and read through

src/rulesets-new/json-api-rules/compound-document-rules.ts
```diff
@@ -0,0 +1,27 @@
...
+import { isOpenApiPath } from "../utils";
+
+export const compoundDocuments = new ResponseBodyRule({
+  name: "disallow compound documents",
+  matches: (responseBody, rulesContext) =>
+    !isOpenApiPath(rulesContext.operation.path) &&
+    ["200", "201"].includes(responseBody.statusCode) &&
+    responseBody.contentType === "application/vnd.api+json",
+  rule: (responseAssertions) => {
+    responseAssertions.body.added.not.matches({
```
niclim, line 11: There's `.not` inverse conditions on all our standard matchers - this should provide the inverse of the matched block.

src/rulesets-new/json-api-rules/pagination-rules.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    return (
+      method === "get" &&
+      !isSingletonPath(rulesContext) &&
+      responseBody.statusCode === "200" &&
+      responseBody.contentType === "application/vnd.api+json"
+    );
+  },
+  rule: (responseAssertions) => {
+    responseAssertions.body.added.matches({
```
niclim, line 73: `matches` is a partial match at the openapi spec node that we are testing.

src/rulesets-new/json-api-rules/resource-object-rules.ts
```diff
@@ -0,0 +1,323 @@
...
+                  type: "string",
+                },
+              },
+            },
+          },
+        },
+      },
+    ];
+
+    responseAssertions.body.added.matchesOneOf(validSchemaShapes);
```
niclim, line 238: This effectively replaces this https://github.com/snyk/sweater-comb/blob/main/src/rulesets/jsonapi.ts#L373 and https://github.com/snyk/sweater-comb/blob/main/schemas/get-post-response-data.yaml

src/rulesets-new/json-api-rules/__tests__/resource-object-rules.test.ts
```diff
@@ -0,0 +1,312 @@
...
+    });
+
+    test("status code 201", () => {
+      const afterJson = {
+        ...baseJson,
+        paths: {
+          "/api/example": {
+            get: {
+              responses: {
+                "201": {
```
cmars, line 72: Hmm... a GET shouldn't respond 201 Created. We might be missing a rule. If this is pre-existing behavior, fine for a followup.

src/rulesets-new/json-api-rules/__tests__/resource-object-rules.test.ts
```diff
@@ -0,0 +1,312 @@
...
+      };
+      const results = ruleRunner.runRulesWithFacts(ruleInputs);
+      expect(results.length).toBeGreaterThan(0);
+      expect(results.every((result) => result.passed)).toBe(true);
+      expect(results).toMatchSnapshot();
+    });
+  });
+
+  describe("valid delete shapes", () => {
+    test("status code 200", () => {
```
cmars, line 236: I prefer the "passes when ..." / "fails when ..." test naming we had in the property rules, because you could have variations on this like:

"passes when a delete responds with 200 and there is a JSON API response body"
"fails when a delete responds with 200 and there is no response body"
"passes when a delete responds with 204 and there is no response body"
"fails when a delete responds with 204 and there is a response body"

src/rulesets-new/json-api-rules/__tests__/status-code-rules.test.ts
```diff
@@ -0,0 +1,158 @@
...
+      ...TestHelpers.createRuleInputs(baseJson, afterJson),
+      context,
+    };
+    const results = ruleRunner.runRulesWithFacts(ruleInputs);
+    expect(results.length).toBeGreaterThan(0);
+    expect(results.every((result) => result.passed)).toBe(false);
+    expect(results).toMatchSnapshot();
+  });
+
+  test("valid delete 2xx codes", () => {
```
cmars, line 44:
```suggestion
  test("fails with invalid delete 2xx codes", () => {
```

src/rulesets-new/json-api-rules/pagination-rules.ts
cmars: This requires a `links` to be declared, but doesn't make any assertions on what kind of links. We'll want to follow up on this one.

src/rulesets-new/json-api-rules/resource-object-rules.ts
```diff
@@ -0,0 +1,323 @@
...
+const responseDataForPatch = new ResponseBodyRule({
+  name: "response data for patch",
+  matches: (responseBody, rulesContext) =>
+    rulesContext.operation.method === "patch" &&
+    responseBody.statusCode === "200" &&
+    responseBody.contentType === "application/vnd.api+json",
+  rule: (responseAssertions) => {
+    responseAssertions.body.added.matches({
+      schema: {
+        properties: {},
```
cmars, line 20: We'll also want to follow up here... there's a bit more structure we'll want to require -- a JSON API resource.

src/rulesets-new/json-api-rules/resource-object-rules.ts
```diff
@@ -0,0 +1,323 @@
...
+          },
+        },
+      },
+    });
+
+    responseAssertions.body.changed.matches({
+      schema: {
+        properties: {
+          jsonapi: {
+            type: Matchers.string,
```
cmars, line 138: Interesting... could I write my own Matcher object?

One that comes to mind is the common LinkProperty type: a oneOf type union.

cmars (commented):
Some questions in this one. It's entirely possible that I'm asking for more rule coverage than the existing rules provided below -- in that case I'd rather land parity with the existing implementation first, before adding more coverage in followups.

Will need to give it a second look tomorrow.

src/rulesets-new/json-api-rules/__tests__/resource-object-rules.test.ts
niclim: It looks like we're missing a similar one to this for GET + PATCH responses https://github.com/snyk/sweater-comb/blob/main/src/rulesets/jsonapi.ts#L56-L74

src/rulesets-new/json-api-rules/resource-object-rules.ts
niclim: From what it seems like, there's a bit of overlap between the rules regarding shape. With the `matches` helper, this should be much easier to write a single rule

src/rulesets-new/json-api-rules/__tests__/resource-object-rules.test.ts
niclim: Makes sense! Updated test cases names

cmars (approved):

PR #271, opened 2022-05-18
chore: migrate specification rules to rule authoring v2

niclim:
Small one this time! Migrating the [specification rules](https://github.com/snyk/sweater-comb/blob/main/src/rulesets/specification.ts) in this PR.

There's one more PR for rule migrations to come, and then we'll need to do a bit of regression testing, then we'll be able to switch over to the new flow.


package.json
```diff
@@ -74,6 +74,7 @@
     "urijs": "^1.19.11"
   },
   "jest": {
+    "maxWorkers": 5,
```
niclim, line 77: @cmars I've been noticing that jest tests occasionally fail with `Call retries were exceeded - ChildProcessWorker.initialize (node_modules/jest-worker/build/workers/ChildProcessWorker.js:193:21)` - [failure here](https://app.circleci.com/pipelines/github/snyk/sweater-comb/683/workflows/1c318698-c280-4249-98b1-b77ad5667253/jobs/782)

Think this is due to: https://github.com/facebook/jest/issues/8769#issuecomment-919811307 - in which case this should limit the parallel workers running in ci

cmars (approved):

PR #272, opened 2022-05-18
feat: woollypully just-in-time script, design doc

cmars:
Starting a Sweater Comb API release governance tool, code-named
"woollypully".

WIP: want to connect this up to an alternative ruleset.
WIP: want to get some design feedback from those more in-the-know on k8s.


src/woollypully/DESIGN.md
```diff
@@ -0,0 +1,69 @@
...
+- Match on those we need to govern (public OpenAPI ones)
+- Apply checks to all the OpenAPI versions provided by those
+
+The main entrypoint in this woollypully sidecar makes use of these endpoints to coordinate the gating.
+
+### Comparing current and proposed
+
+TODO: open problem...
+
+The API discovery isn't exposed outside the pod, so how would we compare with the current running service?
```
cmars, line 25: VU possibly

jlourenc: It shouldn't be difficult to expose the API discovery if needed.

src/woollypully/DESIGN.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Sidecar containers and readiness probes
+
+A WoollyPully sidecar container may be injected into pods. The sidecar discovers the APIs, checks their OpenAPI contracts for compliance, and makes a pass / fail determination.
```
jlourenc, line 29: I'm wondering here if Cerberus shouldn't be responsible for this job with the help of a `WoollyPully` library.

Rationales:
- Cerberus should already be part of the "service" stack
- Cerberus already parses the API discovery and OpenAPI specification
- Cerberus is already responsible for some part of the API governance

mrded:
How will we update such sidecars? let’s say we add more rules into Sweater Comb, we probably want all services to obey them immediately.

cmars:
> How will we update such sidecars? let’s say we add more rules into Sweater Comb, we probably want all services to obey them immediately.

Actually, not in this case. We have to start thinking longer-term with these longer-term maintenance release versions. If something has released to beta or GA, it's already been approved for release and general distribution, likely after significant engineering work to get it there.

Holding that beta or GA release accountable to ever-changing new rules as a precondition for backwards-compatible and incremental updates could become quite disruptive and onerous for teams needing to ship a minor incremental update to support it. Instead, we want to "lock in" the standards that were in place on the release version date.

If a new breaking change version is released, then that new version should absolutely be held to the standards effective on its release (version) date.

So we wouldn't update sidecars at all. What we'd do is, regularly publish sweater-comb / woollypully images with a date-based versioning scheme, which would be used to match API version.

mrded:
I think, having a version of sweater-comb in the openapi spec would be useful so that we know what standards it applies. It will also improve visibility on who needs an update a lot.  That won't block anyone from deploying quick fixes.

cmars:
> I think, having a version of sweater-comb in the openapi spec would be useful so that we know what standards it applies. It will also improve visibility on who needs an update a lot. That won't block anyone from deploying quick fixes.

Because OpenAPI specs are versioned by date, it should be possible to automatically match the most recent version of sweater-comb in effect when the version was release.

This will require a date-based versioning scheme for Sweater Comb releases: tag Sweater Comb / Woollypully releases by date using the same `YYYY-mm-dd` convention. Then matching standards to OpenAPI versions is similar to how we resolve requested -> actual API versions.

src/woollypully/DESIGN.md
jlourenc: > A WoollyPully library might look like a sidecar, since our API linting (this project) is Typescript, and Cerberus is Go. But if Cerberus can orchestrate it, that might be interesting...

Yeah, the orchestration bit is the difficult bit. Maybe there are ways: https://github.com/metacall/golang-typescript-example.

> One property that I'd want here though, that you shouldn't be able to opt out of, or forget to add, or configure WoollyPully. It must be something applied "from the outside" to deployments as policy.

This is also something that we want for Cerberus, being injected instead of being defined by devs.

mrded (approved):

jcsackett (approved):
LGTM; I'm a little confused why we're developing this within sweater-comb; it feels to me more like a vervet situation, where it's a tool that *uses* sweater-comb.

Not enough confusion for me to block on it being developed here, just wondering if it's to speed things up or if there's a specific reason this needs to live in the same repo.

PR #273, opened 2022-05-19
chore: migrate operation rules to rule authoring v2

niclim:
Migrating the [operation rules](https://github.com/snyk/sweater-comb/blob/main/src/rulesets/operation.ts) in this PR.

This should be the last one to migrate. There's a few things we should do before making the switch over to the new rule running API:
- I've got a change coming from optic to provide a way to write custom matcher error messages. Doing a `operationAssertions.requirement.matches({ summary: Matchers.string })` assertion will just show the operation and expected shape, but in this case, you can just add an error message that says operations must have a summary.
- Connect up the existing e2e tests to the new ruleset tests (there appears to be two, [internal JS tests](https://github.com/snyk/sweater-comb/blob/main/src/rulesets/tests/end-end.test.ts) and [full CLI tests](https://github.com/snyk/sweater-comb/tree/main/end-end-tests) - we'll want to run both, and spot check some of the error messages + runs to see if there's anything we want to address here

### DIFF
--- a/src/rulesets-new/operation-rules.ts
+++ b/src/rulesets-new/operation-rules.ts
@@ -1,0 +1,414 @@
+import { OpenAPIV3 } from "@useoptic/openapi-utilities";
+import {
+  RuleError,
+  Ruleset,
+  OperationRule,
+  Matcher,
+  Matchers,
+  ResponseRule,
+} from "@useoptic/rulesets-base";
+import { camelCase, snakeCase } from "change-case";
+import { links } from "../docs";
+import { isBreakingChangeAllowed } from "./utils";
+
+const operationId = new OperationRule({
+  name: "operation id",
+  docsLink: links.standards.operationIds,
+  matches: (operation, ruleContext) => {
+    const changeDate = new Date(ruleContext.custom.changeVersion.date);
+    const effectiveOnDate = new Date("2021-07-01");
+    return (
+      !isBreakingChangeAllowed(ruleContext.custom.changeVersion.stability) &&
+      changeDate > effectiveOnDate
+    );
+  },
+  rule: (operationAssertions) => {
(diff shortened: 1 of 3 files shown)


src/rulesets-new/operation-rules.ts
```diff
@@ -0,0 +1,414 @@
...
+        const name = parameter.value.name;
+        if (name === "group_id" || name === "org_id") {
+          if (!parameter.value.schema) {
+            throw new RuleError({
+              message: "expected parameter to have a schema",
+            });
+          }
+
+          if (
+            !("$ref" in parameter.value.schema) &&
```
cmars, line 186: Please remind me, are refs guaranteed to be resolved? We may easily reuse org and group parameters with $refs.

cmars (approved):
LGTM. Really like the table tests!

src/rulesets-new/operation-rules.ts
niclim: Yes, with a caveat with circular references (circular references will have a `$ref` value)

In the circular reference case, as long as we dereference at _least_ one level, rules will run against the dereferenced values.

Right now, optic-ci doesn't let you use circular references, in the next version release, we'll allow circular references with at least one level dereferenced, so you will always run rules on at least one level deep.

Do you know if there's any case you use (or will use) circular refs. I'm not sure if all the rules here have been written to handle these cases.

cmars: I think we've avoided circular references... that sounds wild. No, we haven't tried that. Not sure we would want to!

Now I'm curious, are there valid and useful applications for circular refs in OpenAPI? I don't believe I've ever seen this in the wild.

niclim: I think there are usecases for graph APIs. It's fairly uncommon to see it in regular REST APIs, but it does happen and we don't want optic-ci to crash when we detect one circular ref!

PR #276, opened 2022-05-24
chore: remove rules v1

niclim:
Removes the old rule v1 code, we switched over to rules v2 in https://github.com/snyk/sweater-comb/pull/275 - we can merge this now, or let the new rules code bake in a little before merging this


cmars (approved):
LGTM 🎉

cmars (approved):
Still LGTM

PR #284, opened 2022-05-27
docs: format request parameter

### LINKED ISSUE #278: Document format query parameter convention in public REST APIs
We're favoring a `?format=` parameter for endpoints which support providing responses in different representations.

Need to document:
- What the `format` parameter means and how to use it according to intended purpose.
- How it affects responses w.r.t. different content types.

cmars:
Adds guidance on how to support alternative media content type
representations of a resource in our REST API standards.

Fixes #278


docs/standards/rest.md
```diff
@@ -115,6 +115,46 @@ headers:
 
 [Header field names are case insensitive](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2). Snyk v3 API specs must use kebab case for consistency. All non-standard headers that are unique to Snyk must begin with `snyk-` (e.g. `snyk-requested-version`).
 
+## <a id="request-parameters"></a>Request Parameters
+
+### <a id="formats"></a>Formats
+
+Some resource types may be expressed with a media content-type format other than JSON API. These alternative formats may be offered with these format request parameters:
```
ianlivingstone, line 122: I'd expect this header to be optional, and that all resources must have a JSON API representation.

mcombuechen: I think the initial reason that this came up, was that for SBOM documents it does not make sense to have a JSON API representation. It’s certainly possible, but @garethr brought up the point that it gives no added value to the consumer. Responses need to be parsed and queried to retrieve the actual content, the JSON API envelop is just a necessary wrapper to get it past validation, but other than that none of the added data from JSON API yield any benefit to the consumer; in fact it makes usage by the consumers slightly more cumbersome. @garethr does this somewhat sum up your concerns?

@ianlivingstone are you concerned that we introduce a backdoor for teams at Snyk to implement services that don’t adhere to JSON API, when in fact they should?

cmars: @ianlivingstone "may be expressed" / "may be offered" in the https://datatracker.ietf.org/doc/html/rfc2119#section-5 sense.

@mcombuechen JSON API enables our consumers to paginate through collections, and navigate a graph of resources through the relationships they declare. It's not really for me to say whether it makes sense to represent SBOM content in some abstract representation in JSON API attributes -- this is certainly a domain-specific judgement call.

If it does not make sense to represent an SBOM _content_ as JSON API -- we should still provide a JSON API response which at a minimum:
- Identifies the resource (`id` and `type`)
- Relates to the supported media type representations of the content, as relationship links.

Here's why:
- Supports aggregating a bulk collection of SBOMs though relationships (give me all the SBOMS for all the projects in org `org_id`) and pagination.
- Supports async API use cases -- imagine a case where producing an SBOM takes a long time. The JSON API response can indicate the operational state of a request for an SBOM, and a link to the created media content type(s) when it's ready.

cmars: Updated taking the above into account:

- A JSON API response is a MUST, with some bare minimum content requirements.
- A complete JSON API representation is a SHOULD, but if it doesn't fit or make sense, just implement the minimum.

I think that covers the above, PTAL.

mcombuechen (approved):

docs/standards/rest.md
```diff
@@ -115,6 +115,55 @@ headers:
...
+
+#### `format` query parameter
+
+This is a URL query parameter of the form `?format=format_name`, where _format_name_ is an application-specific short nickname for the media content-type. This parameter's schema must be defined as an enum.
```
robguinness-snyk, line 126: Is "short nickname" important here, or is the idea to use whatever colloquial name exists for the format (whether it's short or not)? I understand you want to distinguish from the IANA-style content media type here but "short nickname" feels too ambiguous.

docs/standards/rest.md
```diff
@@ -115,6 +115,55 @@ headers:
...
+
+Continuing with the above example, `Accept: application/sarif+json`.
+
+#### Response
+
+The response to a requested format may be:
+
+__200 OK__: The `Content-Type` response header indicates the requested format. The response body contains content in the requested format.
+
+Public REST API endpoints must provide a JSON API response.
```
robguinness-snyk, line 142: Is this in conflict with the above and below? Perhaps:

_Public REST API endpoints must provide a JSON API response, unless another response format is explicitly requested._

docs/standards/rest.md
```diff
@@ -115,6 +115,55 @@ headers:
...
+__400 Bad Request__: The error should indicate an unknown or unsupported format, and may indicate which formats are supported.
+
+__406 Not Acceptable__: The format is supported but this particular resource cannot be represented in the requested format for some application-specific reason. This is not common.
+
+#### Constraints
+
+- Formats are only supported on GET operations of an equivalent JSON API single resource.
+- Formats are not allowed on collections because these must be paginated by JSON API links.
+  - Creation of large artifacts (regardless of media type) may require async API techniques (coming soon).
+- If a resource supports formats:
```
robguinness-snyk, line 160: Perhaps "If a resource supports alternative formats:"

PR #285, opened 2022-05-28
fix: regression in operation naming rules for experimental stability

cmars:
Some of the operation rules ported over to the new DSL were matching
!isBreakingChangeAllowed where they shouldn't have.

The result was a small window where operation ID and parameter naming
rules were not applied to experimental.

Technically this change introduces a tighter constraint which would
normally require a new REST API rules version.

However since this is a regression, which affected only experimental
(which is allowed breaking changes) and the regression was only released
for a small window of several days, the potential impact to fixing this
in-place should be small. Projects will need to conform to these rules
in order to promote to beta anyway.

In the event this change does disrupt API work, affected projects may
pin their Sweater Comb dependency version to buy themselves some time to
update their APIs.

### DIFF
--- a/src/rulesets/rest/2022-05-25/__tests__/end-end.test.ts
+++ b/src/rulesets/rest/2022-05-25/__tests__/end-end.test.ts
@@ -115,6 +115,47 @@ describe("end-end-tests", () => {
     expect(results).toMatchSnapshot();
   });
 
+  it("fails for the right reason with invalid parameters", async () => {
+    const results = await snapshotScenario(
+      undefined,
+      "spec.yaml",
+      resourceDate("repos", "2022-04-04"),
+      {
+        changeDate: "2022-05-27",
+        changeResource: "repos",
+        changeVersion: {
+          date: "2022-04-04",
+          stability: "experimental",
+        },
+        resourceVersions: {},
+      },
+    );
+    expect(results.every((result) => result.passed)).toBe(false);
+    expect(results.filter((result) => !result.passed)).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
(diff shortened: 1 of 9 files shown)


src/rulesets/rest/2022-05-25/__tests__/end-end.test.ts
```diff
@@ -115,6 +115,40 @@ describe("end-end-tests", () => {
...
+        expect.objectContaining({error: 'expected parameter name repoId to be snake case'}),
+        // This is a false positive on a pagination rule triggering. It triggers
+        // because it fails to find a valid trailing path parameter, which it
+        // uses to determine whether the endpoint is a single resource or a
+        // collection. Solving that might be a nice-to-have, but adding special
+        // cases there does not seem appropriate either.
+        //
+        // TODO: The right solution to this might be to make some rules take
+        // precedence over others with a priority score. If a higher scored rule
+        // fails, do not bother showing the lower-scored rules?
```
cmars, line 146: @niclim @acunniffe What do you think about this scenario? Do you see much demand for some rules being able to suppress others to avoid false positives? It's a UX concern -- too much noise and the rule errors aren't actionable, leading to frustrated API developers.

niclim: Interesting, I see. This does feel similar to other hierarchical concerns - for example: "POST endpoints must have response status 201 + shape of `{ id: string }`". The first rule enforces the shape, the second one is scoped it to the first rule.

We could build in a way to link rules, but I worry that the interdependence of rules will make it less clear when rules run.

I think this however, is a common construct (must have "thing", and "thing" must conform to some shape or other rule), so perhaps there's something there we can look into.

PR #287, opened 2022-05-28
fix: do not allow parameters in the path root

### LINKED ISSUE #253: Disallowing path variables as resources
We discovered that https://github.com/snyk/sweater-comb/blob/main/docs/standards.md#resource-collections-are-plural are not actually catching and failing checks when teams are analyzing and building new specs for paths. 

Example:
An endpoint is added `/{var1}/{var2}/{var3}/things` passes checks.

We should disallow the "prefix" or start of a path as a variable since it violates the resource naming and constraints.

cmars:
Do not allow resource paths to declare a parameter in the path root.

This is currently a _de facto_ rule, because API gateway administrators
will not allow catch-all paths at the root like this. This PR makes it
_de jure_.

It's a new constraint, but no standards bump required, because it would
never have been routed in the first place.

Fixes #253


cmars:
CC @genslein

Also @acunniffe @niclim is this a good way to express this with the new DSL? Seemed pretty intuitive to write!

Todai88 (approved):

PR #289, opened 2022-05-31
feat: API governance admission controller webhook

cmars:
Add a Kubernetes admission controller webhook which will inject the
woollypully just-in-time API standards sidecar check into service
deployments.

The webhook doesn't do anything useful yet, it just does enough to confirm that Kubernetes is calling it.


cmars:
Decided to do this in a separate project.

PR #291, opened 2022-06-01
fix: regression in batch POST rule

cmars:
This fixes a regression in the batch POST rule which allows a 204
response when the request body `data` property is an array.

When the rule was ported over, it seems to have changed to expect the
schema of the request body to be an array. It should instead be of the
form:

```
{
  data: [{id: "some-id", type: "some-type", ...}]
}
```

### DIFF
--- a/src/rulesets/rest/2022-05-25/json-api-rules/__tests__/status-code-rules.test.ts
+++ b/src/rulesets/rest/2022-05-25/json-api-rules/__tests__/status-code-rules.test.ts
@@ -115,14 +115,19 @@ describe("status code rules", () => {
     const afterJson = {
       ...baseJson,
       paths: {
-        "/api/users/{user_id}": {
+        "/api/users": {
           post: {
             requestBody: {
               content: {
                 "application/vnd.api+json": {
                   schema: {
-                    type: "array",
-                    items: {},
+                    type: "object",
+                    properties: {
+                      data: {
+                        type: "array",
+                        items: {},
+                      },
+                    },
                   },
                 },
               },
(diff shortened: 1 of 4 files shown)


src/rulesets/rest/2022-05-25/json-api-rules/__tests__/status-code-rules.test.ts
```diff
@@ -115,14 +115,19 @@ describe("status code rules", () => {
     const afterJson = {
       ...baseJson,
       paths: {
-        "/api/users/{user_id}": {
+        "/api/users": {
```
cmars, line 118: This test case makes me realize we might need a rule added -- do not allow POSTs at all on a single resource (path ending with an id).

src/rulesets/rest/2022-05-25/utils.ts
```diff
@@ -10,8 +10,14 @@ export const isBatchPostOperation = (requests) => {
   const request = requests.find(
     (request) => request.contentType === "application/vnd.api+json",
   );
-
-  return request ? request.value.flatSchema.type === "array" : false;
+  const requestSchema = request?.raw.schema;
```
cmars, line 13: @niclim Question about `flatSchema` vs `raw` here... I found `raw` in the debugger, is this the correct way to get access to the OpenAPI document object model? `flatSchema` seemed to drop the properties in the top-level schema object so I couldn't use that. Just want to make sure I'm building on solid ground here. Thanks!

niclim: Oh I see, yeah this is fine. Flatschema is a flattened version of the openapi spec and since you are looking at this from the request level this is ok to do

cmars: Thanks!

PR #293, opened 2022-06-01
fix: required properties must be declared

### LINKED ISSUE #240: required properties must be declared
Declaring required properties that are not actually properties of the object will never be satisfied, so should not be allowed.

cmars:
This should be considered a backward-compatible requirement because an
object schema which declares required properties that are not declared
will never pass request or response validation.

Fixes #240.

### DIFF
--- a/src/rulesets/rest/2022-05-25/property-rules.ts
+++ b/src/rulesets/rest/2022-05-25/property-rules.ts
@@ -185,6 +185,54 @@ const arrayWithItemsInResponse = new ResponseBodyRule({
   },
 });
 
+const requiredPropertiesDeclaredInRequestBody = new RequestRule({
+  name: "request schema properties",
+  rule: (requestAssertions) => {
+    requestAssertions.body.requirement(
+      "declare required properties in objects",
+      (body) => {
+        checkRequiredProperties((body.raw as any).schema);
+      },
+    );
+  },
+});
+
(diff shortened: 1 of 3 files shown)


src/rulesets/rest/2022-05-25/property-rules.ts
```diff
@@ -185,6 +185,54 @@ const arrayWithItemsInResponse = new ResponseBodyRule({
...
+      "declare required properties in objects",
+      (body) => {
+        checkRequiredProperties(body.raw.schema as OpenAPIV3.SchemaObject);
+      },
+    );
+  },
+});
+
+const checkRequiredProperties = (schema: OpenAPIV3.SchemaObject) => {
+  const schemas = [schema];
```
jcsackett, line 213: Since we're only ever passing one schema, and you then operate on schemas independently in the while loop, why make this an array? Seems like it would be simpler to just operate on the passed in schema then wrap it like this.

I'm assuming there's a detail to this that I'm missing which requires an array?

jcsackett (approved):
LGTM, but I'm confused by a detail in the actual checker.

src/rulesets/rest/2022-05-25/property-rules.ts
cmars: Conceptually this is doing a recursive check on all properties which are themselves objects, but an array is being used as a stack here to eliminate that recursion since TS isn't tail-call optimized.

Issue #257, opened 2022-05-03
201 response schema missing "type" when generated with operation workflow


jcsackett:
The output of `sweater-comb workflow operation add-create things` results in a 201 response like:

```
      responses:
        '201':
          description: Created thing successfully
          headers:
            snyk-version-requested:
              $ref: >-
                #/components/x-rest-common/headers/VersionRequestedResponseHeader
            snyk-version-served:
              $ref: '#/components/x-rest-common/headers/VersionServedResponseHeader'
            snyk-request-id:
              $ref: '#/components/x-rest-common/headers/RequestIdResponseHeader'
            snyk-version-lifecycle-stage:
              $ref: '#/components/x-rest-common/headers/VersionStageResponseHeader'
            deprecation:
              $ref: '#/components/x-rest-common/headers/DeprecationHeader'
            sunset:
              $ref: '#/components/x-rest-common/headers/SunsetHeader'
            location:
              $ref: '#/components/x-rest-common/headers/LocationHeader'
          content:
            application/vnd.api+json:
              schema:
                properties:
                  jsonapi:
                    $ref: '#/components/x-rest-common/schemas/JsonApi'
                  data:
                    type: object
                    description: thing resource object
                    required:
                      - id
                      - type
                    properties:
                      id:
                        type: string
                        format: uuid
                        example: d5b640e5-d88c-4c17-9bf0-93597b7a1ce2
                      type:
                        $ref: '#/components/x-rest-common/schemas/Types'
                      attributes:
                        $ref: '#/components/schemas/ThingAttributes'
                      relationships:
                        $ref: '#/components/schemas/ThingRelationships'
                    additionalProperties: false
                  links:
                    $ref: '#/components/x-rest-common/schemas/PaginatedLinks'
```

The `schema` has no `type` field. It should have `type: object`.

JaapRood:
Nice find! These new operations are generated through the use of Workflow Templates, defined right here in sweater comb. They're basically just mutations to an in-memory OpenAPI structure which Optic observes, to then generate and apply the right patches for it. The template for this particular  one lives here:

https://github.com/snyk/sweater-comb/blob/main/src/workflows/templates/operations/create.ts#L28-L70

It, and the other operation workflows, use this bit of template to define the actual bit of schema that has the missing `type` property:

https://github.com/snyk/sweater-comb/blob/ba6aee3548f0e8f6fa37c102fbe5144870550eb9/src/workflows/templates/schemas.ts#L4-L30

PR #295, opened 2022-06-02
fix: adds missing type for response bodies

jcsackett:
Fixes: https://github.com/snyk/sweater-comb/issues/257

### DIFF
--- a/src/rulesets/rest/2022-05-25/json-api-rules/resource-object-rules.ts
+++ b/src/rulesets/rest/2022-05-25/json-api-rules/resource-object-rules.ts
@@ -17,11 +17,13 @@ const responseDataForPatch = new ResponseBodyRule({
   rule: (responseAssertions) => {
     responseAssertions.body.added.matches({
       schema: {
+        type: "object",
         properties: {},
       },
     });
     responseAssertions.body.changed.matches({
       schema: {
+        type: "object",
         properties: {},
       },
     });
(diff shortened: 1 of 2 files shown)


src/rulesets/rest/2022-05-25/json-api-rules/resource-object-rules.ts
```diff
@@ -17,11 +17,13 @@ const responseDataForPatch = new ResponseBodyRule({
   rule: (responseAssertions) => {
     responseAssertions.body.added.matches({
       schema: {
+        type: 'object',
```
cmars, line 20: This is slightly better than what we had before, but what we had wasn't very good. This should be asserting a [valid JSON API PATCH response](https://jsonapi.org/format/#crud-updating-responses) for a 200, similar to how it's done below for GET and POST responses (see `dataProperty` below).

cmars: I wonder if we should refactor a bit -- a rule requiring a JSON API resource response could use a match function that looks at method, response code and content type, to consolidate the decision logic on when a response should contain a resource (200 or 201 for a POST, 200 for GET and PATCH, etc.)

cmars (commented):
Still needs tests.

To really require `type:` for all schemas, it will need to be done recursively, similar to what was done in #240, to ensure all nested schemas in properties have `type:` declared as well.

That should be done in a separate rule (though a schema-walker might be refactored out of #240).

jcsackett:
To make sure we're on the same page, we're good with this fix for the specific issue but you're noting a separate issue that we don't have full validation of type for all schemas. 

Assuming that's a correct understanding I'll land this and create an issue for the recursive aspect.

PR #305, opened 2022-06-09
fix: reserve _at suffix for date-time formatted properties

cmars:
Properties ending in _at must be format: date-time (ISO8601/RFC3339)


src/rulesets/rest/2022-05-25/property-rules.ts
```diff
@@ -133,15 +133,31 @@ const enumOrExample = new RequestRule({
...
+
+const responseDateFormatting = new ResponseBodyRule({
+  name: "response property date formatting",
+  docsLink: links.standards.formats,
+  rule: (responseAssertions) => {
+    responseAssertions.property.added("use date-time for dates", (property) => {
+      if (property.value.key.endsWith("_at")) {
+        if (property.value.flatSchema.format !== "date-time") {
+          throw new RuleError({
+            message: "expected property to have format date-time",
```
jcsackett, line 160: Assuming this is the error message a user sees, it might be more user-friendly to be specific, e.g. `properties ending in _at expected to have format date-time`

jcsackett (approved):

src/rulesets/rest/2022-05-25/property-rules.ts
cmars: Good point... messages should be transparent so it doesn't become a game of "appease the mysterious oracle". I could also link to the documentation on this.

cmars:
This took way more work getting tests to pass than I'd anticipated... 😅

PR #306, opened 2022-06-09
chore: refactor woollypully integration tests and improve image publish

cmars:
Refactor woollypully integration tests to reuse the fake-service for
cases:

- Standalone usage of woollypully in a deployment
- Injection into a pod with the mutating webhook
  (sweater-comb-admissions)

Improve woollypully image publishing to push image tags.


cmars:
Going to JFDI this -- it's mostly about local development (tilt), or CI changes that can only be tested in Circle after it lands.

PR #308, opened 2022-06-12
chore: Add 422 status to common components

vasser:
This is proposal PR to include `422 Unprocessable Entity` in the common components.
Having this status code in the schema will make API responses more flexible and granular.
As long as will help us to avoid overuse of `400 Bad Request`.

As this is draft PR it does not include actual changes in the code and tests, like extending `valid4xxCodes`, for example.

### DIFF
--- a/components/common.yaml
+++ b/components/common.yaml
@@ -406,6 +406,19 @@ responses:
     content:
       application/vnd.api+json:
         schema: { $ref: '#/schemas/ErrorDocument' }
+
+  '422':
+    description: 'Unprocessable Entity: Unable to process the contained instructions of the requested operation.'
+    headers:
+      snyk-version-requested: { $ref: '#/headers/VersionRequestedResponseHeader' }
+      snyk-version-served: { $ref: '#/headers/VersionServedResponseHeader' }
+      snyk-request-id: { $ref: '#/headers/RequestIdResponseHeader' }
+      snyk-version-lifecycle-stage: { $ref: '#/headers/VersionStageResponseHeader' }
+      deprecation: { $ref: '#/headers/DeprecationHeader' }
+      sunset: { $ref: '#/headers/SunsetHeader' }
+    content:
+      application/vnd.api+json:
+        schema: { $ref: '#/schemas/ErrorDocument' }
   
   '500':
     description: 'Internal Server Error: An error was encountered while attempting to process the request.'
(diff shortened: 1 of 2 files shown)


cmars (commented):
Thanks for raising this. This response code shows up in parts of the v1 API, so this was bound to come up. I'm not sure about adding this one to the new REST API though. My thinking so far on this:

[RFC 4918, Section 11.2](https://www.rfc-editor.org/rfc/rfc4918#section-11.2), defines HTTP 422 as "unable to process for some reason other than syntax or content type negotiation" -- such as an application specific reason. It offers an example:

> if an XML request body contains well-formed (i.e., syntactically correct), but semantically erroneous, XML instructions.

Some concerns I have with supporting HTTP 422:

- Rather than rejecting requests on semantics, it'd be better IMO to make it impossible to express invalid requests in the first place. This would make for a clearer and easier to use API. In other words, I think we should be turning the types of errors that would be "semantic 422s" into "syntactical schema 400s" as much as possible in our model.
- This 400 vs 422 distinction may only exist for WebDAV... RFC4918 defines this response as a WebDAV extension to HTTP, designed for WebDAV use cases.
- The [definition of HTTP 400](https://httpwg.org/specs/rfc7231.html#rfc.section.6.5.1) doesn't exclude semantic errors or make this distinction.
- [JSON API error responses](https://jsonapi.org/format/#errors) are capable of expressing business logic errors and making these kinds of situational distinctions and application-specific reasons (syntax, semantics, blocklists, etc).

I'd like to pose some questions that might help us decide the fate of 422, and refine its use if we do decide to adopt it:

- Do we have a strong need to differentiate between an HTTP 400 and a 422?
- How would we expect clients to react differently to these different response codes?
- What would our guiding principles be, that inform our decisions of whether to use a 400 or 422 response in our services?

We will need strong guidance here, so that our errors are consistent across the API.

vasser:
Thanks for the detailed answer.

Your ideas sound clear to me and I do understand the reason why we excluded 422 from the common statuses.

First of all, I do not have a strong opinion on whether we should bring 422 or not. Initially, I thought we should return 400 for syntax errors and 422 for semantics.

Reading your answer, I see all the concerns and understand that 422 is more like an exception, as it kind of sprouts from the 400, right?

Answering your questions as want to understand if we are on the same page and I got your points.

> Do we have a strong need to differentiate between an HTTP 400 and a 422?
- No we do not, on the one hand. On the other - won't we put too much on the shoulders of 400? My concern is that developers can start misusing 400 and just throw more things there

> How would we expect clients to react differently to these different response codes?
- For 400 clients should go and read the documentation if they did something that contradicts what is written there
- For 422 clients should check if they misused combinations of params and request body (for example)

> What would our guiding principles be, that inform our decisions
- Same as I wrote for the first question

PR #309, opened 2022-06-13
feat: meta object standards

cmars:
Provide guidance on the use of meta objects in our API, and refine
property naming rules to allow free-form JSON below the resource meta
object's top-level properties.


carissadean (approved):

jcsackett (approved):
This looks like a good addition to handle circumstances like SCM integrations. Nice!

PR #311, opened 2022-06-15
chore: REST API standards documentation on sparse fieldsets.

cmars:


gablaxian:
It's not explicitly mentioned, but I am assuming we're not following JSON API's type parameter syntax and function, e.g. `attributes[group]=name`

cmars:
We're not. I mention some of the reason why [here](https://github.com/snyk/sweater-comb/blob/main/docs/principles/jsonapi.md#query-parameters-and-json-api) on a different parameter, but maybe it deserves special mention since we're apply a principle of exception to a wider swath of JSON API.

cmars:
When we propose expansion through relationships (another principled departure from JSON API which will come with its own reasoning), I'm thinking we may want sparse fieldsets for those as well. I'm considering a dotted form there (`?relationships.foo.attributes=a,b,c`). Has symmetry with this proposal... it's also a familiar convention (JSONPath).

PR #312, opened 2022-06-16
chore: add error object links per JSON API spec

cmars:
This adds the missing `links` property to the JSON API error object.

See also, [9.2 Error Objects](https://jsonapi.org/format/#error-objects)
in the JSON API specification.

Also proposing this releases as "Common Model 1.0.1" when it lands:
* Merged into branch `common-model-v1`
* Tagged as `common-model-1.0.1`

PR #314, opened 2022-06-20
Added further properties to error link

robh-snyk:
This change extends the definition of `ErrorLink` to include other available error properties, such as self, related, and pagination properties. 

This update ensures the definition of error links in `rest-node-libs` now matches the OpenAPI definition. This slightly extends [this pull request](https://github.com/snyk/sweater-comb/pull/312/files) to use further properties.

### DIFF
--- a/components/common.yaml
+++ b/components/common.yaml
@@ -45,6 +45,12 @@ schemas:
     description: A link that leads to further details about this particular occurrance of the problem.
     properties:
       about: { $ref: "#/schemas/LinkProperty" }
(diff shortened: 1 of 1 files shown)


jcsackett:
@robh-snyk is there still a need for this PR? I think we're still at a point where we don't want these to be expected links for all errors, so adding it to the common definition seems like a misstep to me, but if there's a broader need for it I'd love to hear it.

Otherwise I'd like to close this PR as I'm cleaning up the long list of old open PRs against this project.

PR #315, opened 2022-06-22
chore: documentation on expansion of related resources.

cmars:
Proposal on how Snyk APIs will support API expansion of related
resources in responses.

### DIFF
--- a/docs/principles/jsonapi.md
+++ b/docs/principles/jsonapi.md
@@ -115,7 +115,28 @@ Data objects may declare relationships to other resources — "links with struct
 
 `relationships` is a mapping of "relation name" ⇒ "relation object". That relation object must conform to the structure shown above; it should provide `links.related`, and must provide `data.id` and `data.type`.
 
-Relationships must not provide the actual content of the related resource — it may only link to related resources. Links must also declare a [version]() in the URL to the related resource. This should be the *resolved-version* of the requested resource (not the requested version).
+### <a id="relationships-and-expansion"></a>Relationships and Expansion
+
+Snyk REST APIs support expansion of related resources by enriching relationship [resource identifier objects](https://jsonapi.org/format/#document-resource-identifier-objects) with attributes. Snyk APIs use relationship expansion as an alternative to JSON API [Compound Documents](#rough-compound-documents). This is a Snyk extension to the JSON API specification.
+
+```json
+{
+    "id": "7d1bae82-346a-4f7b-a8cb-37c8f159e415", // unique resource uuid
+    "type": "some-resource", // resource type
+    "attributes": { /* some-resource's attributes */ },
+    "relationships": {
+        "other-resource": {
+            "links": { /* ... */ },
+            "data": {
+                "id": "9f14199e-6330-4b01-a52d-4aaa0ffd29ef", // related entity's ID
+                "type": "other-resource", // related entity's type
+                "attributes": { /* other-resource's attributes */ }
+            },
+        },
+        /* ... */
+    }
+}
+```
 
 ### Pagination and Links
 
@@ -198,19 +219,13 @@ Cursor position identifiers are determined by the links given in a paginated res
 
 ## <a id="the-rough-parts"></a>The Rough Parts
 
-Other parts of the JSON API specification we avoided.
+Parts of the JSON API specification we avoided or intentionally depart from.
 
-### Locating resources through any relationship
+### <a id="rough-compound-documents"></a>Compound documents
 
-JSON API describes how resources may be requested or even modified through their relationships. For example, you might request `resource2/id2` alternatively with `/path/to/resource1/id1/relationships/resource2/id2` if they are related. We don't support or allow this. This is one area where we've intentionally departed from JSON API. We want to provide the links among our resources to form a graph — but leave the actual navigation and traversal of that graph to GraphQL!
+JSON API [compound documents](https://jsonapi.org/format/#document-compound-documents) store all related data objects in an `included` array together, mixing types. An array of `anyOf:` objects can certainly be expressed in JSON Schema, but in practice, this is an awkward data structure to work with, and an unnecessary layer of indirection.
 
-Only `related` links are allowed in relationships.
-
-### Compound documents
-
-As discussed above in *relationships*, related objects are linked, but not included in our responses. This also applies to compound documents in JSON API.
-
-Aside from strategic differences — we're building GraphQL on top of simple resources — we found the implementation of compound documents to be a problem for generating code from OpenAPI descriptions that allow them. JSON API compound documents store all related data objects in an `included` array together, mixing types. An array of polymorphic `anyOf:` objects can certainly be expressed in JSON Schema, but in practice, OpenAPI code generators struggle with processing such schema — especially statically-typed compiled languages.
+Snyk APIs represent expansion more directly by enriching relationship data with attributes.
 
 ### <a id="rough-square-brackets"></a>Square-brackets in query parameters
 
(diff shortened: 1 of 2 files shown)


docs/principles/jsonapi.md
```diff
@@ -115,7 +115,28 @@ Data objects may declare relationships to other resources — "links with struct
...
+Snyk REST APIs support expansion of related resources by enriching relationship [resource identifier objects](https://jsonapi.org/format/#document-resource-identifier-objects) with attributes. Snyk APIs use relationship expansion as an alternative to JSON API [Compound Documents](#rough-compound-documents). This is a Snyk extension to the JSON API specification.
+
+```json
+{
+    "id": "7d1bae82-346a-4f7b-a8cb-37c8f159e415", // unique resource uuid
+    "type": "some-resource", // resource type
+    "attributes": { /* some-resource's attributes */ },
+    "relationships": {
+        "other-resource": {
+            "links": { /* ... */ },
```
mmols-snyk, line 129: Thanks for thinking through our cases - looking forward to our meeting on Monday. 

I was curious about `links` here - would it be appropriate to define a relationship to a resource that does not yet have an API in the `relationships` dict?  An example is `rule_result` in our model - which is something we might eventually have an API for, but we won't at GA. Right now, we've put those details in the attributes dict.

Is there a case for a relationship with no link?

docs/standards/rest.md
```diff
@@ -201,6 +201,44 @@ The `version` URL query parameter, `?version=version_string` is reserved for sel
...
+
+The parameter should be declared in OpenAPI with the same representation as other array parameters, `{"style": "form", "explode": false}`.
+
+The default expansion behavior when `expand` is not specified is "no expansion". If the default expansion differs, this behavior must be documented in the parameter's description.
+
+#### Constraints
+
+Resources may support expansion of any one-to-one relationship. They must not expand one-to-many relationships _in relationship data_. The reason for this is complexity around pagination -- collections must be paginated in Snyk APIs, and this becomes complicated with multiple collections within resources, or collections within collections.
+
+It is still possible to represent a one-to-many relationship with expansion, however. A resource may expand relationships in each resource in a collection response.
```
ianlivingstone, line 220: This isn't super clear. It'd be helpful for there to be explicit examples of one-to-one and one-to-many and how you'd approach those use cases.

docs/standards/rest.md
```diff
@@ -228,6 +266,12 @@ When the `attributes` parameter is not specified, the assumed default behavior i
 
 Aside from the [difference in parameter naming](../principles/jsonapi.md#rough-square-brackets), all other [JSON API requirements and restrictions on sparse fieldsets](https://jsonapi.org/format/#fetching-sparse-fieldsets) apply.
 
+#### Sparse fieldsets and expansion
+
+Sparse fieldsets may be expressed on related expansions by prefixing the parameter with the relationship name and a dot `.`, _relationship_`.attributes`.
+
+For example: `/orgs/{org_id}/projects?expand=target&attributes=name&target.attributes=name` would expand the projects response with related target resources, and only include the `name` attribute in each.
```
ianlivingstone, line 273: Can you expand more than one thing?

docs/standards/rest.md
```diff
@@ -201,6 +201,44 @@ The `version` URL query parameter, `?version=version_string` is reserved for sel
...
 
+### <a id="expansion"></a>Expansion
+
+A resource's relationships may inline the related resource's attributes to conserve API requests and provide a richer response. This inline enrichment of related data attributes is called _expansion_.
+
+#### `expand` query parameter
+
+The `expand` URL query parameter is reserved for expressing expansion of relationships. Resource endpoints are not required to support this parameter, but when they do, they must declare the `expand` parameter schema as an array of enum values. The enum values must match the relationship names provided in the resource data `relationships` map.
+
+The parameter should be declared in OpenAPI with the same representation as other array parameters, `{"style": "form", "explode": false}`.
```
ianlivingstone, line 212: Can we get a rule to enforce this? If it's not in code, it's not really a standard!

docs/standards/rest.md
cmars: Yes.

docs/standards/rest.md
cmars: Yes, in a followup.

docs/standards/rest.md
cmars: I think I need a better canonical concrete example of resources with this kind of relationship. JSON API's `authors` & `articles` would be a lot clearer than `thing` and `this-other-thing` :)

cmars: Updated

docs/standards/rest.md
cmars: Updated

cmars:
@ianlivingstone Updated from your feedback, PTAL. Another thing I'm considering for followup is how we might express filtering through relationships as a general pattern. There I'm thinking of a extending the convention for [User-defined filters](https://github.com/snyk/sweater-comb/blob/main/docs/standards/rest.md#filters):

* The filter parameter name must match the corresponding resource attribute property name
* OR it must match a related resource attribute property, of the form {relationship name}_{attribute name}

This enables more expressive queries like "give me all the critical issues for projects tagged 'foo'".

docs/standards/rest.md
```diff
@@ -201,6 +201,128 @@ The `version` URL query parameter, `?version=version_string` is reserved for sel
...
+}
+```
+
+```
+GET /authors/5c4b6ec5-9ca0-4b14-8ab1-23ee26684cea?expand=books
+
+HTTP 400 Bad Request
+Content-Type: application/vnd.api+json
+{
+  "errors": [{ /* can't do this */ }]
```
jcsackett, line 322: Do we want to error on this? Or just not expand things that we don't support expanding? I get "explicit better than implicit" but I can also see this making things more complicated on the consumer side where you just want to expand everything that can be expanded.

But perhaps that sort of aggro API consumption isn't a thing we should support, as I think about it?

jcsackett (commented):
LGTM, raising one rambling question about errors on bad expansion parameters, which you can feel free to address or ignore 😝

jcsackett (approved):

docs/standards/rest.md
cmars: I think we'd 400 here specifically because you'd be specifying an invalid enum value for `?expand=` if you tried to do this. So you'd basically be dealing with a validation error at this point.

Sweater Comb rules would also prevent it from happening in API design in the first place. In a followup, we can add a rule that does not allow a relationship object's `.data` property to be an array.

cmars: "Expand all" is interesting. I think it'd be fine as an optional shorthand, but I'd like to gauge demand before adding it to the spec.

If you're building UI components for expanded resources, I think you're probably having to integrate specifically with each expansion, so it might be fine to expect those to be enumerated.

I also didn't want to require all relationships to support expansion. In some cases it might be expensive or complicated to expand, or to do so with consistency guarantees, without necessary supporting infrastructure. So this gives the resource owner some breathing room to link first, then support inlining when it's needed, and release to beta / GA when it can be made performant / reliable.

TerenceSnyk: This seems pretty limiting for an obvious use case, the author-books example is illustrative.  Our users will want to see projects->targets, groups->orgs, orgs->users, etc.  Many of the obvious use cases require one->many expansion so not supporting that out of the gate feels not great.

cmars: I think it is supported, you just need to "flip it around" so that you're expanding items in a paginated collection.

So in your examples, I think you could:

- `GET /orgs/{org_id}/projects?expand=target` to get a collection of projects with their target expanded in each item
- `GET /orgs?expand=group` to get a collection of orgs with their group expanded
- `GET /users?expand=org` to get a collection of users with their org expanded

The only use case this doesn't support is a case where you want to deal with multiple collections simultaneously, such as an N:M relationship. This becomes really difficult to express pagination over however.

PR #316, opened 2022-06-23
chore: document collection counts metadata & parameters

cmars:
Defining a `meta` object convention for representing collection counts,
and query parameters to request counts on collections.

### DIFF
--- a/docs/standards/rest.md
+++ b/docs/standards/rest.md
@@ -165,7 +165,55 @@ headers:
 
 ### Meta objects
 
-Meta objects should be used sparingly. Snyk's data model representation of a resource must be represented in JSON API data attributes. Alternative representations of that resource in other formats should use format and content type negotiation instead.
+#### <a id="meta-collection-counts"></a>Collection counts
+
+Resource collection responses may express summary statistics in a `meta` object, using the following optional properties:
+
+```yaml
+type: object
+properties:
+  count:
+    type: number
+    minimum: 0
+  count_by:
+    type: object
+    properties: {}
+    additionalProperties:
+      type: object
+      properties: {}
+      additionalProperties:
+        type: number
+        minimum: 0
+```
+
+What this might look like by example (a total count, and counts grouped by two different attributes):
+
+```
+{
+  "meta": {
+    "count": 10,
+    "count_by": {
+      "color": {
+        "red":  2,
+        "blue": 3,
+        "green": 5
+      },
+      "t_shirt_size": {
+        "S": 2,
+        "M": 3,
+        "L": 3,
+        "XL": 2
+      }
+    }
+  }
+}
+```
+
+Collection counts may be added to the top-level `meta` object in a resource collection response, or in a relationship `meta` object, when the relationship is "to-many".
+
+#### Custom metadata
+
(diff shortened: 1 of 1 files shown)


docs/standards/rest.md
```diff
@@ -197,6 +245,40 @@ The following parameters are reserved for specific purposes in our API. These na
...
+If `meta_count` is not specified but `meta_count_by` is specified, `meta_count=only` is implied. The following table represents the default behaviors of different combinations of these parameters and their resulting affect on the response contents:
+
+| `meta_count` value | `meta_count_by` value | Response has `.meta.count` | Response has `.meta.count_by` | Response has `.data[]` |
+|------------|---------------|--------|-|-|
+| _not set_ | _not set_ | ❌ | ❌ | ✅ |
+| _not set_ | _one or more group(s)_ | ✅ | ✅  | ❌ |
+| `only` | _*_ | ✅ | _*_ | ❌ |
+| `with` | _*_ |  ✅ | _*_ | ✅ |
+
+An endpoint may choose to change this default behavior. If so, it must be documented in the parameter(s) description(s).
```
mitechie, line 280: Should this be a stronger suggestion in the standard? I'm thinking how in v1 there are some APIs that support pagination and some that don't. I almost wonder if the ideal is that some level of the resource tree should have a default meta_count functionality supported.

jcsackett (commented):
This seems fine, but I'm wondering if we are having too many options where "this is the default, if you're doing something else you must document it"; maybe we need to start saying "do it this way"?

Do we have known cases already where folks have a need to do something else?

docs/standards/rest.md
cmars: Obtaining the total count and grouped counts is likely to be a separate query that has to be made, and possibly an index scan to get an exact count. For a sufficiently large collection, you might cache this, or estimate counts in a HyperLogLog structure instead (but that's another thing to then store and keep in sync). This makes me hesitant to declare it a MUST across the board for all resources up front, before we know what we're in for here.

The things we have going for us here though, that v1 lacks:

- The OpenAPI spec will still declare whether a given resource collection supports counts, and how to get it.
- Customer demand can then drive improvement in those specs which do not have counts, with new version releases.

We could decide later that API version releases after a certain date must include counts. I think that would need an ADR for the reasons above. If we have that though, and we're confident that services can handle it, I'd feel comfortable that we could roll out such a constraint for all API versions releasing after that decision date.

cmars: It seems reasonable to make the _behavior_ standard though... "if you implement counts, this is how it works". Just wanted to separate out these different threads.

cmars: Shored up the wording on this one. @jcsackett had similar concerns, I'm convinced.

cmars:
> This seems fine, but I'm wondering if we are having too many options where "this is the default, if you're doing something else you must document it"; maybe we need to start saying "do it this way"?
> 
> Do we have known cases already where folks have a need to do something else?

I think you're right. Better to provide strong direction here -- and a rule that can be automated, than leave it too open. Updated!

jcsackett (approved):
Thanks for the updates @cmars!

PR #317, opened 2022-06-30
fix: allow breaking removals in a spec being removed

jcsackett:
When a spec is sunset it can be removed; the lifecycle checks are done
on the compiled version to ensure that the spec can be removed, but
sweater-comb complains about the removal of properties and operations as
part of the spec being deleted.

Since there's a check to ensure it's safe for the spec to be removed,
this commit allows the removal of operations and properties as part of
the spec deletion.

### DIFF
--- a/src/rulesets/rest/2022-05-25/utils.ts
+++ b/src/rulesets/rest/2022-05-25/utils.ts
@@ -61,3 +61,7 @@ export const isResourceMetaProperty = (property: Field): boolean => {
     ) !== null;
   return isResourceMetaProperty;
 };
+
+export const specIsRemoved = (spec): boolean => {
+  return spec.change === "removed";
+};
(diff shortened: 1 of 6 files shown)


src/rulesets/rest/2022-05-25/utils.ts
```diff
@@ -61,3 +61,7 @@ export const isResourceMetaProperty = (property: Field): boolean => {
     ) !== null;
   return isResourceMetaProperty;
 };
+
+export const specIsRemoved = (spec): boolean => {
+  return spec.change === "removed";
```
cmars, line 66: Does this apply to the entire specification being deleted? Or the entity being evaluated?

jcsackett: The spec being passed in is the spec itself. The whole thing has to be deleted for it to be ok for breaking change deletes to happen, since the lifecycle management of the spec is handled in another rule.

cmars (approved):

PR #324, opened 2022-07-01
docs: updates filtering for expanded values

jcsackett:


docs/standards/rest.md
```diff
@@ -237,6 +237,12 @@ A filter parameter may support a single value to match, or a set of multiple val
 
 Changing a parameter's schema from a single value to multiple values is a non-breaking API change. The inverse however (changing from supporting multple values to a single value) is a breaking API change.
 
+#### Filtering through relationships
+
+A filter parameter may support a value in an [expanded relationship](#expansion). When filtering on values in expanded relationships, the filter is express in the form ?property_name=relationship_name.value_name. For example, using if a query were filtering on a resource, `book`, which had a relationship to `author`, the query could filter on the author's name with `?property_name=author.name`.
```
cmars, line 242:
```suggestion
A filter parameter may support a value in an [expanded relationship](#expansion). When filtering on values in expanded relationships, the filter is expressed in the form `?property_name=relationship_name.value_name`. For example, using if a query were filtering on a resource, `book`, which had a relationship to `author`, the query could filter on the author's name with `?property_name=author.name`.
```

cmars (approved):

cmars:
Does this only support expansion, or could it be used to filter through relationships _in general_? I think this convention would be generally useful for that purpose as well. It seems orthogonal to expansion.

cmars:
Also consider that this will require a parameter naming rule change. The dot `.` will likely be non-compliant.

I have mixed feelings about relating this to expansion. On one hand, if we relate this to expansion, it's possible to write a rule that checks whether the relationship property filter matches the related resource's attributes.

We could start with coupling them together like this, and then break that coupling later if / when we run into a situation that requires it. Easier to relax a standard later than introduce tighter constraints later. And, we might get a clearer idea of where it makes sense to do that.

jcsackett:
> Also consider that this will require a parameter naming rule change. The dot `.` will likely be non-compliant.

I don't think it is; dot notation in query parameters is allowed, and having just tested adding a query param with name `author.name` into a spec in go-service-sample, vervet/sweater-comb don't complain.

jcsackett:
>Does this only support expansion, or could it be used to filter through relationships in general? I think this convention would be generally useful for that purpose as well. It seems orthogonal to expansion.

This has been related to expansion b/c that's the context in which it was proposed; I suppose I can see a case where we don't care about the expanded aspect, but it feels like weird ergonomics to me if I'm filtering on a value and then can't see that value in the output.

But I suppose there's an equally valid argument that it's awkward if I don't say expand and then all my results are expanded.

I'm ambivalent on this--if we want to remove it, I'm fine with that.

jcsackett:
As I've mulled it over, I agree that it's orthogonal to expansion. I've updated it to remove references to expansion.

Another thought I had was if we want to support 1:N or enforce strictly to 1:1 as we do in expansion--I can see a case where we want to filter on a many, but we can't expand the many relationship. Right now it's left unsaid, but we probably want to be explicit one way or another--thoughts?

cmars (approved):

cmars:
> As I've mulled it over, I agree that it's orthogonal to expansion. I've updated it to remove references to expansion.

Sounds good. I don't think this prevents a rule from matching the expansion attributes with filters either -- the rule can still be "if expand is supported on a relationship, filter properties that are declared must match".

> Another thought I had was if we want to support 1:N or enforce strictly to 1:1 as we do in expansion--I can see a case where we want to filter on a many, but we can't expand the many relationship. Right now it's left unsaid, but we probably want to be explicit one way or another--thoughts?

Filtering through an 1 : N relationship is interesting... could such a filter function as a `WHERE EXISTS` subquery on the relationship? Not necessarily something to worry about or define until we have a use case, but could be useful!

jcsackett:
> Filtering through a 1 : N relationship is interesting... could such a filter function as a `WHERE EXISTS` subquery on the relationship? Not necessarily something to worry about or define until we have a use case, but could be useful!

Yeah, I'm content to leave it undefined until someone asks how it should be done, but there's room for interesting query formats there.

PR #328, opened 2022-07-20
feat: sweater-comb lint command

cmars:
Lint command which frontends optic-ci's bulk-compare, using the
project's .vervet.yaml to derive the base ref and glob patterns.

### DIFF
--- a/end-end-tests/api-standards/.vervet.yaml
+++ b/end-end-tests/api-standards/.vervet.yaml
@@ -1,0 +1,8 @@
+linters:
+  some-linter:
+    original: main
+apis:
+  some-api:
+    resources:
+      - path: resources
+        linter: some-linter
--- a/src/__tests__/lint.test.ts
+++ b/src/__tests__/lint.test.ts
@@ -1,0 +1,83 @@
+import * as os from "os";
+import * as uuid from "uuid";
+import { lintAction } from "../lint";
+
+// Tests which spawn subprocesses seem to take longer to execute in CircleCI
+const testTimeout = 30000;
+
+describe("lint command", () => {
+  let pushd: string;
+  beforeEach(async () => {
+    pushd = process.cwd();
+    process.chdir(__dirname + "/../..");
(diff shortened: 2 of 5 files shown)


carissadean:
How would this be used within a service? Do you have any examples of the command a user would need to add?

cmars:
> How would this be used within a service? Do you have any examples of the command a user would need to add?

This is intended to deprecate and replace `vervet lint`, so that we have a clearer separation of responsibilities. `vervet` should be responsible only for compiling OpenAPI resource versions, and `sweater-comb` should be responsible for authoring and validating OpenAPI resource specs.

`lint` works like this:

```
$ node build/index.js lint --help
Usage: index lint [options] [path] [base]

Lint OpenAPI files found at the given path. When no arguments are given, OpenAPI resources defined in .vervet.yaml are linted.

Arguments:
  path        API resource path
  base        Base git branch for comparison (default: "main")

Options:
  -h, --help  display help for command
```

```
$ node build/index.js lint ./end-end-tests/api-standards/resources                                                                                  Running bulk-compare with globs ./end-end-tests/api-standards/resources/**/[2-9][0-9][0-9][0-9]-[0-1][0-9]-[0-3][0-9]/spec.yaml
Found file end-end-tests/api-standards/resources/repos/2022-04-04/spec.yaml
Bulk comparing 1 comparisons
Comparing main:end-end-tests/api-standards/resources/repos/2022-04-04/spec.yaml to /Users/c/Projects/sweater-comb/end-end-tests/api-standards/resources/repos/2022-04-04/spec.yaml

   PASS  Specification


   PASS  GET /orgs/{orgId}/repos/{repoId}


   PASS  DELETE /orgs/{orgId}/repos/{repoId}


   PASS  GET /orgs/{orgId}/repos


   PASS  POST /orgs/{orgId}/repos


0 changes detected
645 checks passed
0 checks failed
```

carissadean (approved):

PR #329, opened 2022-07-28
feat: allows arrays with oneOf/anyOf/allOf definitions to pass lint

jcsackett:


src/rulesets/rest/2022-05-25/__tests__/property-rules.test.ts
```diff
@@ -867,6 +867,207 @@ describe("body properties", () => {
...
+                  description: "",
+                  content: {
+                    "application/json": {
+                      schema: {
+                        type: "object",
+                        properties: {
+                          things: {
+                            type: "array",
+                            items: {
+                              oneOf: [],
```
cmars, line 971: Should we pass if it's an _empty_ composite type though? I think we should require at least one concrete type in here.

I suppose the worst pathological case would be nested composite types with nothing at the bottom -- so to be _really_ comprehensive we should do a recursive descent to make sure all "leaf-node" type-paths are defined (not empty composites). This would be the most formally-correct solution.

For example, this type contains a whole lot of choices, all leading to "no type defined":

```
items:
  oneOf:
    - oneOf: []
    - {}
    - anyOf:
        - allOf: []
        - {}
```

A helper function `isCompletelyDefinedType` could probably tackle this.

A band-aid quick-fix solution might be to check that "all composites must contain one or more simple non-empty type" and deal with the formal solution in a followup. AFAIK no one's trying to nest composites at the moment (but it may happen unintentionally later!)

src/rulesets/rest/2022-05-25/property-rules.ts
```diff
@@ -209,9 +209,14 @@ const arrayWithItemsInRequest = new RequestRule({
...
-              message: "type was not found array items",
-            });
+            const oneOf = property.raw.items["oneOf"];
+            const allOf = property.raw.items["allOf"];
+            const anyOf = property.raw.items["anyOf"];
+            if (!oneOf && !allOf && !anyOf) {
+              throw new RuleError({
+                message: "type was not found array items",
+              });
+            }
```
cmars, line 219: Recommend turning this into a helper, for incremental improvement as described above. Asserting a non-empty type seems like a generally useful thing we may want for property types as well as array item types.

src/rulesets/rest/2022-05-25/__tests__/property-rules.test.ts
jcsackett: I was under the impression that there were other checks for composite types having concrete types--if not, wouldn't it be better to have those checks across all schemas rather than making that part of this check?

cmars: There might be... I'll have to look (it's been a little while)

jcsackett: Did a quick check and there aren't; I'm going to file an issue for validating that composites are completely typed, and go with a band-aid helper in this branch.

jcsackett: https://github.com/snyk/sweater-comb/issues/330

cmars (approved):

PR #331, opened 2022-07-29
fix: recursively check for fully-defined array types

cmars:
Support nested composite types when checking that an array type is fully
defined.

### DIFF
--- a/src/rulesets/rest/2022-05-25/utils.ts
+++ b/src/rulesets/rest/2022-05-25/utils.ts
@@ -70,22 +70,44 @@ export const specIsRemoved = (spec): boolean => {
 export const isFullyTypedArray = (
   array: OpenAPIV3.ArraySchemaObject,
 ): boolean => {
-  if (!array.items || !("type" in array.items)) {
-    for (const key of ["oneOf", "allOf", "anyOf"]) {
-      const composite = array.items[key];
-      if (composite) {
-        if (composite.length === 0) {
-          return false;
-        }
-        for (const obj of composite) {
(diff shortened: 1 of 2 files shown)


src/rulesets/rest/2022-05-25/utils.ts
```diff
@@ -70,22 +70,41 @@ export const specIsRemoved = (spec): boolean => {
...
-        if (composite.length === 0) {
-          return false;
-        }
-        for (const obj of composite) {
-          if (!("type" in obj)) {
-            return false;
-          }
-        }
-        return true;
+  return isFullyTypedType(array.items as OpenAPIV3.SchemaObject);
```
jcsackett, line 73: Doesn't need to be fixed now, but items can also be OpenAPIV3.ReferenceObject, which only has `$ref`, at which point I think we just declare it handled elsewhere. Same for all composite types.

jcsackett (approved):

src/rulesets/rest/2022-05-25/utils.ts
cmars: I think references are guaranteed to be completely resolved when rules evaluate, even though the type signature allows it. Will check though.

cmars: Yes.. you'll see a similar cast in other rules. https://github.com/cmars/sweater-comb/blob/53aa718b5c92ccf80e957a3ffba659aee29cb30d/src/rulesets/rest/2022-05-25/property-rules.ts#L259 for example

PR #337, opened 2022-08-15
feat: add meta_count query parameter component

jgresty:
meta_count is defined in our standards (docts/standards/rest.md) to
specify how we return the meta.count object in the response. This is
useful for clients that only need a count of the objects and not the
objects themselves.


cmars (approved):
Thanks for this!

PR #338, opened 2022-08-15
chore: bump dependency versions

cmars:
Fix several known issues in parse-url and other deps.


jcsackett (commented):
LGTM, but this really throws into relief the weirdness of having a yarn and npm in this repo 😁

jcsackett (approved):

PR #347, opened 2022-08-18
fix: pin sweater-comb commander to compatible version

acunniffe:
@jcsackett reported an issue where yarn installs were leading to internal errors. I was able to reproduce this locally and found that the issue was in `commander` package. Optic uses a different minor version than Snyk and a recent change on our end led to conflicting node_modules. Yarn and NPM handle those conflicts differently and that's why we're seeing different behavior between the two. 

There are changes coming to custom rule setups that should insource more of these concerns to the SDKs and lead to fewer dependencies in your project. Pinning should keep everything consistent for the time being :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "chai": "^4.3.4",
     "chalk": "^4.0.0",
     "change-case": "^4.1.2",
-    "commander": "^9.4.0",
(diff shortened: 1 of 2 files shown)


cmars (approved):

cmars:
Thanks for looking into this!

PR #348, opened 2022-08-18
fix: allows dotted notation in query params

jcsackett:
Our standards use `relationship_name.property_name=value` and
`property_name.sub_property=value` for filtering on properties and
relationships. The current snake case rules catch these as rule violations,
when they're valid.

This updates the query param rules to check for snake case on all segments of
a dotted notation path, but doesn't flag the `.` as invalid naming.

### DIFF
--- a/src/rulesets/rest/2022-05-25/operation-rules.ts
+++ b/src/rulesets/rest/2022-05-25/operation-rules.ts
@@ -14,6 +14,16 @@ import {
   isCompiledOperationSunsetAllowed,
 } from "./utils";
 
+export const validDottedName = (name) => {
+  return (
+    !name.startsWith(".") &&
+    !name.endsWith(".") &&
+    name.split(".").every((n) => {
+      return n !== "" && snakeCase(n) === n;
+    })
+  );
+};
+
 const operationId = new OperationRule({
   name: "operation id",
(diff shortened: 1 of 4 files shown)


src/rulesets/rest/2022-05-25/operation-rules.ts
```diff
@@ -141,7 +141,11 @@ const parameterCase = new OperationRule({
       "use the correct case",
       (queryParameter) => {
         const name = queryParameter.value.name;
-        if (snakeCase(name) !== name) {
+        if (
+          name.split(".").every((n) => {
+            snakeCase(n) === n;
+          })
+        ) {
```
cmars, line 148: I'd suggest defining a function for this and unit testing it. I think snake_case on each split is good, but we probably also want to make sure we don't allow consecutive, leading or trailing dots (no `.foo`, `foo..bar`, `foo.bar.` for example).

cmars (approved):
LGTM with a suggestion

src/rulesets/rest/2022-05-25/operation-rules.ts
jcsackett: That's a good call, thanks!

PR #355, opened 2022-09-15
fix: require singleton GET responses to conform to JSON API

cmars:
Singleton resource GET responses should be required to conform to JSON API resource object requirements.

Added test coverage for this missed rule condition with a bit of refactoring to reuse test data.

### DIFF
--- a/end-end-tests/api-standards/test-bulk.bash
+++ b/end-end-tests/api-standards/test-bulk.bash
@@ -14,15 +14,17 @@ trap "rm -rf $tempdir" EXIT
 set +e
 
 function assert_ok {
-    if [ "$?" != "0" ]; then
-        echo "unexpected rc=$?"
+    rc=$?
+    if [ "$rc" != "0" ]; then
+        echo "unexpected rc=$rc"
         exit 1
     fi
 }
 
 function assert_err {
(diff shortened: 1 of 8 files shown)


end-end-tests/api-standards/resources/thing/2021-11-10/001-singleton.yaml
```diff
@@ -21,7 +21,7 @@ paths:
       parameters:
         - { $ref: "#/components/x-rest-common/parameters/Version" }
       responses:
-        "201":
+        "200":
```
carissadean, line 24: Non-blocking (especially as it's not part of this PR) but the description and operationId above and below should be updated to match that this is a get request

carissadean (approved):

end-end-tests/api-standards/test-bulk.bash
```diff
@@ -14,15 +14,17 @@ trap "rm -rf $tempdir" EXIT
 set +e
 
 function assert_ok {
-    if [ "$?" != "0" ]; then
-        echo "unexpected rc=$?"
```
cmars, line 18: This `$?` was wrong, it was the status of the test above! That didn't affect the decision logic, but the log output was very wrong.

PR #359, opened 2022-09-19
chore: add 422 error

DoronElir:
Hi, we don't have a 422 error, so I can't use it on my repo

I took the description from wiki https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#cite_note-RFC_4918-17

### DIFF
--- a/components/common.yaml
+++ b/components/common.yaml
@@ -427,6 +427,19 @@ responses:
     content:
       application/vnd.api+json:
         schema: { $ref: '#/schemas/ErrorDocument' }
+        
+  '422':
+    description: 'Unprocessable Entity: The request was well-formed but was unable to be followed due to semantic errors'
+    headers:
+      snyk-version-requested: { $ref: '#/headers/VersionRequestedResponseHeader' }
+      snyk-version-served: { $ref: '#/headers/VersionServedResponseHeader' }
+      snyk-request-id: { $ref: '#/headers/RequestIdResponseHeader' }
+      snyk-version-lifecycle-stage: { $ref: '#/headers/VersionStageResponseHeader' }
+      deprecation: { $ref: '#/headers/DeprecationHeader' }
+      sunset: { $ref: '#/headers/SunsetHeader' }
+    content:
+      application/vnd.api+json:
+        schema: { $ref: '#/schemas/ErrorDocument' }
   
   '500':
     description: 'Internal Server Error: An error was encountered while attempting to process the request.'


cmars (changes requested):
I really appreciate the contribution to our standards here -- something I would like to continue to encourage.

Unfortunately in this case, I do not think we should use HTTP 422 in our REST API.

I think I could be convinced to allow 422 if there were clear guidelines and rules we could put in place explaining when to use 400 vs 422, to ensure they are consistently applied, and to explain them well to our users. Even better, if a 422 meaningfully improves API developer experience in some way.

Why I'm currently deciding against 422:

The HTTP status code should instruct the client on matters of protocol. There is no useful difference between a 400 Bad Request and 422 Unprocessable Entity from an _HTTP protocol perspective_; in both cases, you made a bad request and you should not try it again.

While it is useful to explain to the caller the reason why the request is bad, the distinction between what constitutes a syntactical vs semantic error may not always be completely clear. I have concerns that we may struggle to use 400 vs 422 status codes appropriately and consistently across our services, and this may lead to inconsistency in our API.

It is better from a devex perspective to make semantic errors impossible to express with well-defined schemas. JSON schema unfortunately cannot express some of these constraints (maybe OpenAPI 4 will use something more powerful like CueLang! :D). Does that then mean these violated constraints are syntactical errors which our OpenAPI is not yet capable of expressing yet, or semantic errors due to the limitations of JSON schema? This makes me feel like 400 vs 422 is possibly an internal application implementation detail.

To our users, it does not matter -- the request is bad and they need to know why. So there is a vitally important difference between the two kinds of errors from an application perspective though! We can and should make this distinction between syntax and semantics at the application level in our internal error catalog (#proj-error-catalog).

Less status codes here makes the API easier to understand from a user perspective: a 400 with a JSON API error explaining why the request was bad. We can still make the distinction between syntactic and semantic errors in application error responses if that is useful to our users, by adding these application-level cases to our error catalog.

DoronElir:
> I really appreciate the contribution to our standards here -- something I would like to continue to encourage.
> 
> Unfortunately in this case, I do not think we should use HTTP 422 in our REST API.
> 
> I think I could be convinced to allow 422 if there were clear guidelines and rules we could put in place explaining when to use 400 vs 422, to ensure they are consistently applied, and to explain them well to our users. Even better, if a 422 meaningfully improves API developer experience in some way.
> 
> Why I'm currently deciding against 422:
> 
> The HTTP status code should instruct the client on matters of protocol. There is no useful difference between a 400 Bad Request and 422 Unprocessable Entity from an _HTTP protocol perspective_; in both cases, you made a bad request and you should not try it again.
> 
> While it is useful to explain to the caller the reason why the request is bad, the distinction between what constitutes a syntactical vs semantic error may not always be completely clear. I have concerns that we may struggle to use 400 vs 422 status codes appropriately and consistently across our services, and this may lead to inconsistency in our API.
> 
> It is better from a devex perspective to make semantic errors impossible to express with well-defined schemas. JSON schema unfortunately cannot express some of these constraints (maybe OpenAPI 4 will use something more powerful like CueLang! :D). Does that then mean these violated constraints are syntactical errors which our OpenAPI is not yet capable of expressing yet, or semantic errors due to the limitations of JSON schema? This makes me feel like 400 vs 422 is possibly an internal application implementation detail.
> 
> To our users, it does not matter -- the request is bad and they need to know why. So there is a vitally important difference between the two kinds of errors from an application perspective though! We can and should make this distinction between syntax and semantics at the application level in our internal error catalog (#proj-error-catalog).
> 
> Less status codes here makes the API easier to understand from a user perspective: a 400 with a JSON API error explaining why the request was bad. We can still make the distinction between syntactic and semantic errors in application error responses if that is useful to our users, by adding these application-level cases to our error catalog.

@cmars Hi, thanks for the verbose replay
while I don't see the introduction of the 422 error code adding that much complexity, I understand what you are saying and can agree with it.

for now, I'll change my use case to use 400 error code

PR #362, opened 2022-09-22
fix: allow resource IDs to be format: uri

cmars:
Some resources are uniquely located by a structured identifier, and do not have a UUID. Examples include packages (identified by a pURL), issues, and CVEs.

For IDs which are not UUIDs, allow the use of a URI to identify the resource. Even if the URI is of a custom type, requiring a URI creates the right incentive to define a well-formed structure for that identifier.

### DIFF
--- a/docs/standards/rest.md
+++ b/docs/standards/rest.md
@@ -2,7 +2,21 @@
 
 In order to provide a consistent [API as a platform](../principles/api_program.md), Snyk APIs have additional requirements, building on [JSON API](../principles/jsonapi.md) and [Versioning](../principles/version.md) standards.
 
-## Organization and group tenants for resources
+# Resource IDs
+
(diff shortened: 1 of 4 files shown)


cmars:
Some caveats considering when reviewing this:
* URI IDs will be escaped when used in path or query parameters. So a resource for a pURL-identified package `pkg:bitbucket/birkenfeld/pygments-main@244fd47e07d1014f0aed9c` would look like `/packages/pkg:bitbucket%2Fbirkenfeld%2Fpygments-main@244fd47e07d1014f0aed9c`
* A resource might support multiple ID types, a `oneOf:` of `format: uuid` or `format: uri`
* `format: uri` isn't enough to validate a client-originated ID in a POST request. This PR may need to restrict that for security reasons. 
* We're putting more trust in the service owner to use appropriate and well-designed URIs

carissadean (approved):

PR #364, opened 2022-10-14
feat: lower pagination limit to 1

eliecharra:
Hi team 👋🏻 

I have an usecase where I want to retrieve only the latest created item of a collection, sorted by creation date. 
The current Limit specification does not actually allow us to fetch only one item.
I do think that it could make a lot of sense to allow that.

Also, I removed the `multipleOf: 10`, not sure if this is compatible with `limit=1`.

### DIFF
--- a/docs/principles/jsonapi.md
+++ b/docs/principles/jsonapi.md
@@ -213,7 +213,7 @@ Our API uses these reserved parameters for pagination:
 
 - `starting_after` - Return `limit` records after the record identified by cursor position `starting_after`.
 - `ending_before` - Return `limit` records before the record identified by cursor position `ending_before`.
-- `limit` - Size of page, in increments of 10, up to 100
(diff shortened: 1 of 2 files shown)


docs/principles/jsonapi.md
```diff
@@ -213,7 +213,7 @@ Our API uses these reserved parameters for pagination:
 
 - `starting_after` - Return `limit` records after the record identified by cursor position `starting_after`.
 - `ending_before` - Return `limit` records before the record identified by cursor position `ending_before`.
-- `limit` - Size of page, in increments of 10, up to 100
+- `limit` - Size of page, in increments of 1, up to 100
```
jcsackett, line 216:
```suggestion
- `limit` - Number of records to return, up to 100
```

jcsackett (approved):
This looks good, thanks.

PR #368, opened 2022-11-15
feat: adds bulk patch rules

jcsackett:


src/rulesets/rest/2022-05-25/json-api-rules/__tests__/resource-object-rules.test.ts
```diff
@@ -2,11 +2,70 @@ import { OpenAPIV3 } from "@useoptic/openapi-utilities";
 import { RuleRunner, TestHelpers } from "@useoptic/rulesets-base";
 import { context } from "../../__tests__/fixtures";
 import { resourceObjectRules } from "../resource-object-rules";
-
 const baseJson = TestHelpers.createEmptySpec();
 
 describe("resource object rules", () => {
   describe("valid PATCH requests", () => {
+    test("passes when bulk PATCH request body is of the correct form with id format %s and response status code %s", () => {
```
cmars, line 9: Was there an intent to `test.each` here? If not I'd replace with:
```suggestion
    test("passes when bulk PATCH request body is of the correct form with id format uuid and response status code 204", () => {
```

cmars (approved):
Minor suggestion on the test description, otw LGTM!

PR #385, opened 2023-01-09
Revert "chore: bump optic to 0.35.8"

clwiseman:
Reverts snyk/sweater-comb#377

This is a temporary fix revert a recent change in Optic which is
causing issues with some of our existing API specs.


DarrellMozingo (approved):
LGTM!

nit - can you update the commit message to point out this is to temporarily revert a recent optic ci bump which is causing issues with some of our specs?

https://github.com/opticdev/optic/pull/1410

PR #408, opened 2023-03-30
Upgrade Optic to 0.42.4

niclim:
Upgrades Optic to 0.42.4 which includes:
- automatic historic documentation
  - stats, verified documentation, changelog between historic APIs
- new changelogs

Note that this will contain breaking changes - suggested to major version bump here (along side a [PR to remove workflows](https://github.com/snyk/sweater-comb/pull/409)). The majority of the changes in this repo are related to just updating the internal tests / call commands

## Notable changes
- New way to call `optic` - `compare` becomes `diff` and `bulk-compare` becomes `diff-all`
  - `diff` is called in two ways `diff file1 file2` or `diff file --base $gitref`
  - `diff-all` scans the current working directory, pulling all specs with `x-optic-url`, and 
    - e.g. `diff-all --compare-from main` - compares all specs with `x-optic-url` to the main branch
  - to also run rules, append `--check` to the end of these commands
- To register `x-optic-url` you can run `sweater-comb api add` - requires a user to login, but this will sync the API spec to Optic cloud
  - to catch specs that haven't been added with `x-optic-url`, we can add a flag to the CI diff to fail if we detect any unadded specs

## To do following outside of this
- [x] Create a new `snyk` organization in optic - https://app.useoptic.com and invite other users
- [ ] Update vervet to stop calling `optic bulk-compare` and instead call `sweater-comb lint` [here](https://github.com/snyk/vervet/blob/main/internal/linter/optic/linter.go#L152)
- [ ] Update the circleCI orb that runs Optic to run on both `push` and `PR` events ([example snippet here](https://gist.github.com/nalanj/ca9f0be83e6980e0959edbd0fd74c659))
  - check the vars are set `OPTIC_TOKEN` and `GITHUB_TOKEN` - which I think should be the same as before
  - we'll also probably want to add the `--fail-on-untracked-openapi` flag and `--match ${resourceDir}/**/[2-9][0-9][0-9][0-9]-[0-1][0-9]-[0-3][0-9]/spec.yaml`
    - depending on what the circleCI orb calls, this might need to be updated in this PR
- [ ] For each new repo that we upgrade
  - add `x-optic-url` to all openapi specs - `sweater-comb api add` - you'll want to add this to the snyk organization created above
  - bump to use the new sweater-comb version + use new circleCI script
- [ ] Update the developer docs for
  - how to run optic diff / diff-all
  - how to add new apis to optic (optic api add)
  - how to run workflows (removed in https://github.com/snyk/sweater-comb/pull/409)


end-end-tests/api-standards/resources/repos/2022-04-04/spec.yaml
```diff
@@ -1,5 +1,6 @@
 openapi: 3.0.3
 x-snyk-api-stability: experimental
+x-optic-url: https://app.useoptic.com/organizations/orgid/apis/apiId
```
niclim, line 3: I've added a "fake" optic url here since these are just used for test cases, we don't need to upload anything but need them to be picked up by optic cloud

src/index.ts
```diff
@@ -98,9 +109,8 @@ const main = async (): Promise<void> => {
 
   workflowCommand.addCommand(operationCommand);
   cli.addCommand(workflowCommand);
-  cli.addCommand(createLintCommand());
 
-  await cli.exitOverride().parseAsync(process.argv);
+  cli.exitOverride().parse(process.argv);
```
niclim, line 113: Wasn't really sure why but the 9 -> 10 commander update the parse async was capturing the exit code, this should be equivalent though

src/rulesets/rest/2022-05-25/spectral.ts
```diff
@@ -0,0 +1,25 @@
+import {
+  Spectral,
+  RulesetDefinition as SpectralRulesetDefinition,
+} from "@stoplight/spectral-core";
+import { oas } from "@stoplight/spectral-rulesets";
+import { SpectralRule } from "@useoptic/rulesets-base";
+
+const spectral = new Spectral();
```
niclim, line 8: This is just the spectral ruleset pulled out to be a first class citizen (rather than a configurable object)

gablaxian (approved):

jgresty:
Rebased in #410

PR #411, opened 2023-04-05
fix: correct bundled dependencies after update

jgresty:
The bundled dependencies list was not updated to match the updated dependency list.


package.json
```diff
@@ -66,9 +65,8 @@
   },
   "bundledDependencies": [
```
niclim, line 66: I think you'll also want to specify `@stoplight/spectral-core` and `@stoplight/spectral-rulesets` since they include versions of ajv - and the [original commit](https://github.com/snyk/sweater-comb/commit/968ae01fad106be7b7e53ff21f6a581ce15b5b31) that added this in was to address `ajv` version mismatches

gablaxian (approved):

PR #412, opened 2023-04-14
fix: align query version format with vervet

jlourenc:
Vervet, which is used by backends to vet query versions, does:
- accept "ga" as stability
- not accept query versions without a date
- not accept "work-in-progress" as stability

### DIFF
--- a/components/common.yaml
+++ b/components/common.yaml
@@ -213,12 +213,12 @@ schemas:
   QueryVersion:
     type: "string"
     description: "Requested API version"
-    pattern: "^(wip|work-in-progress|experimental|beta|((([0-9]{4})-([0-1][0-9]))-((3[01])|(0[1-9])|([12][0-9]))(~(wip|work-in-progress|experimental|beta))?))$"
+    pattern: "^((([0-9]{4})-([0-1][0-9]))-((3[01])|(0[1-9])|([12][0-9]))(~(wip|experimental|beta|ga))?)$"
 
   ActualVersion:
     type: string
     description: Resolved API version
-    pattern: "^((([0-9]{4})-([0-1][0-9]))-((3[01])|(0[1-9])|([12][0-9]))(~(wip|work-in-progress|experimental|beta))?)$"
+    pattern: "^((([0-9]{4})-([0-1][0-9]))-((3[01])|(0[1-9])|([12][0-9]))(~(wip|experimental|beta|ga))?)$"
 
 headers:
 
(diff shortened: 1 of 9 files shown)


components/common.yaml
```diff
@@ -213,12 +213,12 @@ schemas:
   QueryVersion:
     type: "string"
     description: "Requested API version"
-    pattern: "^(wip|work-in-progress|experimental|beta|((([0-9]{4})-([0-1][0-9]))-((3[01])|(0[1-9])|([12][0-9]))(~(wip|work-in-progress|experimental|beta))?))$"
+    pattern: "^((([0-9]{4})-([0-1][0-9]))-((3[01])|(0[1-9])|([12][0-9]))(~(wip|experimental|beta|ga))?)$"
```
jlourenc, line 216: Maybe we don't want to promote `ga` as a stability and rely on it being the default?

docs/principles/version.md
```diff
@@ -139,14 +139,6 @@ GET /path/to/resource?version=2021-09-21
 
 would match only resource versions released as GA, as of 2021-09-21.
 
-As a shortcut, you can also request "today's latest experimental" with:
-
-```json
-GET /path/to/resource?version=experimental
```
jlourenc, line 145: My little testing on versioning looks to me this is not (no longer?) supported. Is it a legacy behaviour that predates date resolution?

cmars (commented):
If this change gets rolled out service-by-service, I am concerned users not providing the date in the query version will inconsistently see 200s vs 400s across the API.

docs/principles/version.md
cmars: It turns out that the node implementation allows this, but it probably shouldn't.

It's useful to allow OpenAPI spec lookup without the date, e.g. `/openapi/experimental`. In that case, you're asking for the latest version of the OpenAPI spec at experimental stability. You might want to see that, just to get a snapshot of what the latest API looks like. The spec will have annotations telling you at which version each endpoint was introduced -- a nice way to discover the API.

Allowing the date to "float" in other API endpoint requests puts more risk onto the API consumer. If an API consumer builds their software against `?version=experimental`, that can easily break. By not allowing that to happen in the first place, we're removing that risk and promoting responsible and correct use of the versioning system from the start.

components/common.yaml
cmars: That was the original idea. Both the Node and Go implementations of the version format should interpret "no stability specified" as "GA by default".

cmars (commented):
I think this might be fine, if released in such a way that existing specs which link the common components do not introduce an unplanned constraint. I would consider it a breaking change.

jgresty:
Stale

PR #423, opened 2023-05-24
feat: add rules for POST/PATCH/DELETE on relationships

gablaxian:
This adds:
 - Request data rules for `POST`/`PATCH`/`DELETE` requests on `/relationships/` paths.
 - A specific status code rule for `POST` relationship requests, in an effort to avoid overloading either of the existing `POST` status code rules.

The `jsonapi` spec [distinguishes](https://jsonapi.org/format/#crud-updating-resource-relationships) modifications to a resource's relationships from modifications to the resource itself, with different request data requirements. Namely, `attributes` are not required but an `id` is always required.

### DIFF
--- a/src/rulesets/rest/2022-05-25/json-api-rules/__tests__/resource-object-rules.test.ts
+++ b/src/rulesets/rest/2022-05-25/json-api-rules/__tests__/resource-object-rules.test.ts
@@ -173,6 +173,47 @@ describe("resource object rules", () => {
       expect(results.every((result) => result.passed)).toBe(true);
     });
 
+    test("passes when POST request body for a relationship is of the correct form", async () => {
+      const afterJson = {
+        ...baseJson,
+        paths: {
+          "/api/example/relationships/example": {
+            post: {
+              responses: {}, // not tested here
+              requestBody: {
+                content: {
+                  "application/vnd.api+json": {
+                    schema: {
+                      type: "object",
+                      properties: {
+                        data: {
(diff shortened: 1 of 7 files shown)


love-bhardwaj (approved):
I am not sure how urgent this is 😅 LGTM! If not too urgent, I would appreciate it if someone with more Sweater Comb knowledge can have a look.

src/rulesets/rest/2022-05-25/json-api-rules/__tests__/resource-object-rules.test.ts
```diff
@@ -121,6 +121,111 @@ describe("resource object rules", () => {
         expect(results.every((result) => result.passed)).toBe(true);
       },
     );
+
+    test.each(["uuid", "uri"])(
+      "passes when PATCH request body for a relationship is of the correct form",
```
jgresty, line 126: nit: I cannot tell what the "correct form" is unless I read the body of the test, so I don't know if the test is accurate. I'll only comment this once to reduce noise

jgresty (approved):

src/rulesets/rest/2022-05-25/json-api-rules/__tests__/resource-object-rules.test.ts
gablaxian: What would be your recommended improvement here? Pull out the object from within the test and place it in a variable at the start of the test block?

PR #444, opened 2023-07-27
chore: removes woolly pulley deployment

gablaxian:
Woolly Pulley was an aspirational plan for API governance which, sadly, was not able to reach its full potential, due to a lack of ownership. It has been removed from Argo and so does not run within our infrastructure. I am opting to keep in the `src` files and design docs so that maybe one day it can come back. It's not a bad idea, just one which needs support we don't currently have.

Drive by update: remove Swimm doc files.


jgresty (approved):
All the red makes me happy

PR #461, opened 2023-08-14
fix: remove attributes from required post body properties

jgresty:
The upstream JSON:API spec states that only type is required. Requiring attributes means we will be unable to create resources that have no attributes.

An example of what this is blocking is creating app_install resources which have bodies similar to:

    {
        "type": "app_install",
        "relationships": {
            "app": { "data": { "type": "app", "id": 1234 } }
        }
    }

This is perfectly valid under JSON:API specs but fails the current linting due to the missing attributes.

### DIFF
--- a/src/rulesets/rest/2022-05-25/json-api-rules/resource-object-rules.ts
+++ b/src/rulesets/rest/2022-05-25/json-api-rules/resource-object-rules.ts
@@ -93,9 +93,6 @@ const matchPostRequest = {
       type: {
         type: Matchers.string,
       },
-      attributes: {
-        type: "object",
-      },
     },
   },
 };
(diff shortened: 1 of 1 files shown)


love-bhardwaj:
Adding for reference: https://jsonapi.org/format/#crud-creating

love-bhardwaj (approved):
I don't see any issue with it, since the Open API spec allows it.

ajbeattie-snyk (approved):
We should probably update the `Bulk Post` operation as well:

https://github.com/snyk/sweater-comb/blob/main/src/rulesets/rest/2022-05-25/json-api-rules/resource-object-rules.ts#L171

cmars (commented):
Is there a way to say "attributes" is optional, but if it's there, it needs to be an object?

jgresty:
> Is there a way to say "attributes" is optional, but if it's there, it needs to be an object?

I did try to do this but it doesn't look like there is a way

PR #488, opened 2023-09-13
fix: add nix result to path

jgresty:
This package builds @snyk/sweater-comb, which means the sweater-comb binary is not on the exported path from the resulting nix derivation. This is fine when running the result directly as it is picked up from mainProgram - however when consuming into a devShell it means that the binary is not in the users path.


ajbeattie-snyk (approved):

love-bhardwaj (approved):
Don't know much about Nix 🤷 But I trust you with this.

PR #512, opened 2023-10-20
docs: add docs for numerical filters

yeforriak:
Add docs for range numerical filters.

### DIFF
--- a/docs/standards/rest.md
+++ b/docs/standards/rest.md
@@ -249,6 +249,9 @@ To declare this format on a timestamp filter, use `type: string, format: date-ti
 
 The `_before` and `_at_or_before` filters cannot be used together, nor can `_after` and `_at_or_after`. Otherwise filters can be mixed (e.g. `_before` can be used with either `_after` or `_at_or_after`). Doing so implies searching within a range, requiring the `_before` (or its inclusive version) to be an earlier time than `_after` (or its inclusive version).
 
+#### Numerical filters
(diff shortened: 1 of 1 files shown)


jgresty (approved):
It might be difficult to explain this in our public documentation since support will be different per query parameter. If this is something you are going to support in your endpoint it may be worth laying down some recommendations around that area too.

jgresty:
Also an example would be highly appreciated

yeforriak:
> Also an example would be highly appreciated

yes definitely!. I added a quick example:
```
For example, the `count` property can also be used as a range filtering value via `count_gt` or `count_lt` for `count` values greater than or less than the provided filter values
```
is it clear? should I rephrase it? or change the format?

jgresty:
> is it clear? should I rephrase it? or change the format?

That is clear, I was thinking about an example query, eg:
```http
GET https://example.com/foos?count_lt=5
```

yeforriak:
> > is it clear? should I rephrase it? or change the format?
> 
> That is clear, I was thinking about an example query, eg:
> 
> ```httpspec
> GET https://example.com/foos?count_lt=5
> ```

Gotcha, yes good idea. I just added it. Thanks

yeforriak:
> It might be difficult to explain this in our public documentation since support will be different per query parameter. If this is something you are going to support in your endpoint it may be worth laying down some recommendations around that area too

yes, that is a good idea. 
I didn't start with the implementation yet, but as soon as I make some progress I can follow up with a different PR and share some recommendations.

PR #527, opened 2023-11-21
feat: add link formatting to rest standard

DarrellMozingo:
See also: https://github.com/snyk/rest-api-docs/pull/144

### DIFF
--- a/docs/standards/rest.md
+++ b/docs/standards/rest.md
@@ -36,6 +36,21 @@ To declare this format on a timestamp attribute property, use:
 
 `type: string, format: date-time`
 
+### Links
+
+All links should:
+
+* be relative, including `/rest` (customers will need to prepend their instance domain)
+* _not_ include a version parameter (to avoid feature mismatch betwen versions, so a customer will need to include this parameter)
+
+Therefore, a `next` link to an org's project listing might be:
+
+```json
+"links": {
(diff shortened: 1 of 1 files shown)


jgresty (approved):

docs/standards/rest.md
```diff
@@ -36,6 +36,21 @@ To declare this format on a timestamp attribute property, use:
 
 `type: string, format: date-time`
 
+### Links
+
+All links should:
+
+* be relative, including `/rest` (customers will need to prepend their instance domain)
+* _not_ include a version parameter (to avoid feature mismatch betwen versions, so a customer will need to include this parameter)
```
ajbeattie-snyk, line 44: Would it make sense for `self`/`first`/`next` links to include the `version` that was included in the original request? There are already some APIs that do this (e.g. new issue API) so removing it would be an unexpected change.

DarrellMozingo: Yea, I think that's a valid route! My worry is consistency - is that helpful, or more annoying that consumers then need to know when/if to add version, vs always or never?

re breaking change: 💯, both of these are breaking changes depending how endpoints implement today (which is all over the board!). I'm thinking we:
1. Get consensus on what we want to do (this PR)
2. Lint the outcome of this PR somehow
3. Move existing APIs to follow this, which will be a long tail of customer migration

The best time to start is yesterday, second best is today 😄

ajbeattie-snyk: Sgtm!

I don't have strong feelings one way or another - migration will be potentially painful either way :D

Issue #472, opened 2023-08-29
Latest version 2.1.7 won't pass lint on spec changes already released


cmars:
Typically sweater-comb should only apply most Optic rules to the delta between the current working copy and the original reference branch, which is usually `main`.

However, Sweater Comb 2.1.7 is failing on a previously released OpenAPI spec. That makes upgrading the linter impossible for this project, which is currently linted with 2.1.0.

It appears that optic is seeing some of the existing specs as "new", even though they are in git history.

This appears to be a regression that crept in somewhere between 2.1.0 and 2.1.7; 2.1.0 does not have this problem.

This potentially blocks us from being able to upgrade to make use of #471 if/when that lands & releases.

Thanks!

PR #587, opened 2024-02-16
fix: update optic to 0.54.4

corneliu-petrescu:
Update useoptic dependencies to fix axios vulnerability

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "@stoplight/spectral-rulesets": "^1.18.1",
     "@useoptic/optic": "^0.43.0",
     "@useoptic/json-pointer-helpers": "^0.53.16",
-    "@useoptic/openapi-io": "^0.53.15",
+    "@useoptic/openapi-io": "^0.53.16",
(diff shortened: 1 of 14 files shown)


jgresty:
How confident are we that https://github.com/snyk/sweater-comb/issues/472 has been addressed with this release?

package.json
```diff
@@ -53,21 +53,23 @@
...
     "find-parent-dir": "^0.3.1",
     "fs-extra": "^10.0.0",
     "yaml": "^2.3.4",
     "yargs": "~17.6.0"
   },
   "overrides": {
-    "postman-collection": "4.2.0"
+    "postman-collection": "4.2.0",
+    "pac-resolver": "7.0.1",
+    "socks": "2.7.3"
```
jgresty, line 72: Don't need these overrides, the packages that have these listed as dependencies do not point to exact versions. Since these versions have already have entries in our package lock there is no need to explicitly declare overrides, and having them could hurt dependency resolution in the future

jgresty (approved):

PR #591, opened 2024-03-13
Feat/sweatercomb v2 fixes

corneliu-petrescu:
Added tests for:
   - optic diff-all works without x-optic url
   - vervet.yaml excludes is honored by the lint commmand.
   
   
  * DO NOT MERGE *

### DIFF
--- a/end-end-tests/api-standards/resources/thing/2021-11-10/005-baseline-no-optic-url.yaml
+++ b/end-end-tests/api-standards/resources/thing/2021-11-10/005-baseline-no-optic-url.yaml
@@ -1,0 +1,296 @@
+openapi: 3.0.3
+x-snyk-api-stability: experimental
+info:
+  title: v3
+  version: 3.0.0
+servers:
+  - url: https://api.snyk.io/v3
+    description: Public Snyk API
+tags:
+  - name: Thing
+    description: Short description of what Thing represents
+paths:
+  /orgs/{org_id}/thing:
+    post:
+      summary: Create a new thing
+      description: Create a new thing
+      operationId: createThing
+      tags:
+        - Thing
+      parameters:
+        - { $ref: "#/components/x-rest-common/parameters/Version" }
+        - { $ref: '#/components/parameters/OrgId' }
+      responses:
+        "201":
+          description: Created thing successfully
+          headers:
+            snyk-version-requested:
+              {
+                $ref: "#/components/x-rest-common/headers/VersionRequestedResponseHeader",
+              }
+            snyk-version-served:
+              {
+                $ref: "#/components/x-rest-common/headers/VersionServedResponseHeader",
+              }
+            snyk-request-id:
+              {
+                $ref: "#/components/x-rest-common/headers/RequestIdResponseHeader",
+              }
+            snyk-version-lifecycle-stage:
+              {
+                $ref: "#/components/x-rest-common/headers/VersionStageResponseHeader",
+              }
+            deprecation:
+              {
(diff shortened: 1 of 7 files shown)


src/lint.ts
```diff
@@ -99,12 +99,77 @@ export const lintAction = async (
...
+  }
+};
+
+export const createLintPRCommand = () => {
+  const command = new Command("lint-pr")
+    .addArgument(
+      new Argument("base", "base git branch for comparison").argRequired(),
+    )
+    .addArgument(
+      new Argument("to", "base git branch for comparison").argRequired(),
```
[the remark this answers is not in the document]
corneliu-petrescu: Typo, will fix it.

src/lint.ts
```diff
@@ -99,12 +99,77 @@ export const lintAction = async (
...
+      await expectGitBranch(base);
+      await bulkCompare(
+        path.join(path.relative(topDir, vervetConfDir), resource.path),
+        base,
+        to,
+        resource.excludes,
+      );
+    }
+  }
+};
```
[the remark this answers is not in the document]
corneliu-petrescu: The reason I created the second command is because we were using unnamed parameters and there is some very specific logic dependant on the first param: "resource". 
This conflicts with what we're trying to do in the PR workflow, where resource param is not used.

Do you know if there is a way to configure commander to accept something like:  sweater-comb lint base=sha1 to=sha2 and not rely on order?

corneliu-petrescu:
> What is the workflow for end users? Do they run `sweater-comb lint` locally then the orb runs `lint-pr`?

Yes, that was the idea.

src/lint.ts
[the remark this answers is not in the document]
corneliu-petrescu: I refactored the code to reuse the existing lint command

src/lint.ts
```diff
@@ -113,6 +118,15 @@ export const createLintCommand = () => {
         .argOptional()
         .default(defaultBranchName),
     )
+    .option(
+      "--compare-to <compare-to>",
+      "the head ref to compare against. Defaults to the current working directory",
```
jgresty, line 123: How can a ref default to the current working directory?

src/lint.ts
```diff
@@ -113,6 +118,15 @@ export const createLintCommand = () => {
...
+    )
+    .option(
+      "--compare-from <compare-from>",
+      "the base ref to compare against. Defaults to HEAD~1",
+      "HEAD~1",
```
jgresty, line 128: I think head of main is a better default than `HEAD~1`. People typically make several broken commits before running lint

jgresty (approved):
Nice having one entrypoint makes this a lot easier to consume. We might want to think about hiding all the optic commands too since they don't really fit into our workflow, it just adds a bit of confusion if no teams are using them.

src/lint.ts
corneliu-petrescu: This is how the argument is documented in the underlying optic diff-all command.

src/lint.ts
corneliu-petrescu: I updated

corneliu-petrescu:
> Nice having one entrypoint makes this a lot easier to consume. We might want to think about hiding all the optic commands too since they don't really fit into our workflow, it just adds a bit of confusion if no teams are using them.

Removing the optic commands, will require some more investigation
I think right now we need the optic cli init to set setting custom rulesets.
Happy to try to fix it post sweater-comb release.

PR #600, opened 2024-03-21
fix: improve logging

corneliu-petrescu:
Improved logging for skipped endpoints.
Currently, we only log the resource name and not the actual API path


tinygrasshopper:
Are we logging the api key yo stdout?

PR #601, opened 2024-03-22
feat: revert to node18

corneliu-petrescu:
Revert to node18 and semantic-release 22

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "/schemas"
   ],
   "engines": {
-    "node": ">=20"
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "/schemas"
   ],
   "engines": {
-    "node": ">=20"
+    "node": ">=16"
```
jgresty, line 12: Is there something that needs node 16? Lets at least stick to versions that aren't EOL

jgresty (approved):

PR #605, opened 2024-04-01
fix: preserve order of compareTO overrides base

corneliu-petrescu:
bug fix for preserving correct overrides of base and compareTo arguments

### DIFF
--- a/src/lint.ts
+++ b/src/lint.ts
@@ -97,10 +97,17 @@ export const lintAction = async (
         );
         continue;
       }
+
+      /*
+       * base can be override by compareFrom
+       * if neither base nor compareFrom are given use defaultBranchName
+       */
       const base =
         options?.compareFrom ??
+        branchName ??
         linter["optic-ci"]?.original ??
         defaultBranchName;
+
       await expectGitBranch(base);
       await bulkCompare(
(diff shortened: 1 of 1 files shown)


src/lint.ts
```diff
@@ -97,10 +97,17 @@ export const lintAction = async (
...
         continue;
       }
+
+      /*
+       * base can be override by compareFrom
+       * if neither base nor compareFrom are given use defaultBranchName
+       */
       const base =
         options?.compareFrom ??
+        branchName ??
```
jgresty, line 107: We shouldn't have two ways of passing the same data. Either remove it from options or remove the `branchName` param and always use the value from options.

corneliu-petrescu: I see your point. 
I think I'll keep the base param and just remove the compareTO in that case.
Reason: compatiblity with "sweater-comb lint" command

I'll updatete the orb as well

jgresty (approved):
From this PR it isn't obvious why we are removing `--compare-from`, I know it is to simplify the interface but labelling it under "fixes" is not helpful for anyone in the future who looks back at this.

corneliu-petrescu:
I'll update the description to accurately match what it's doing.

PR #623, opened 2024-06-04
fix: relock transitive deps

corneliu-petrescu:


package-lock.json
```diff
@@ -13136,6 +13136,20 @@
       "integrity": "sha512-OO0pH2lK6a0hZnAdau5ItzHPI6pUlvI7jMVnxUQRtw4owF2wk8lOSabtGDCTP4Ggrg2MbGnWO9X8K1t4+fGMDw==",
       "dev": true
     },
+    "node_modules/fsevents": {
```
jgresty, line 13139: Not sure where this came from, none of the other dependencies have introduced it

jgresty (approved):

package-lock.json
corneliu-petrescu: It's marked as an optional dependency. I removed it, reran npm install. It did not get added back

PR #627, opened 2024-07-08
Prevent experimental stability for new endpoints

jgresty:
There have been a lot of experimental endpoints published which then do
not get promoted into a higher stability level. To avoid compounding the
problem we have made the decision to prevent any new experimental
endpoints and instead have teams release more stable endpoints by
default.

This patch introduces a new rule to prevent such endpoints and updates
the documentation to remove references to experimental.

This will be a major version bump so it will not break existing workflows until we roll out the new major version.


src/rulesets/rest/2022-05-25/lifecycle-rules.ts
```diff
@@ -106,7 +106,35 @@ const followSunsetRules = new SpecificationRule({
   },
 });
 
+const noNewExperimental = new SpecificationRule({
```
tinygrasshopper, line 109: How can we feature flag this change? Can we apply a pivot date config here that is read from a github source, so that we can control the rollout based on a the product comms?

jgresty: Is that needed when this is a major version change?

RaduPetreTarean (approved):

PR #644, opened 2024-09-03
Fix: Allow for one to one relationship requests

jac-moran:
As discussed in https://snyk.slack.com/archives/C040Q5G1Q3G/p1725293953086929 , it was detected that all relationship updates are to-many requests, but a relationship request with just an entity should be allowed as well.

This change adds support for that, so that both bodies are allowed in a relationship update (post / patch / delete) request:
* this (as previously):
```
  data: {
    type: "array",
    items: {
      type: "object",
      properties: {
        type: {
          type: Matchers.string,
        },
        id: {
          type: "string",
          format: resourceIDFormat,
        },
      },
    },
  }
``` 
* but also this (single entity):
```
  data: {
    type: "object",
    properties: {
      type: {
        type: Matchers.string,
      },
      id: {
        type: "string",
        format: resourceIDFormat,
      },
    },
  }
```
Tests have been modified as the error message changed, and also positive tests for the single entity cases have been added.

### DIFF
--- a/src/rulesets/rest/2022-05-25/json-api-rules/__tests__/resource-object-rules.test.ts
+++ b/src/rulesets/rest/2022-05-25/json-api-rules/__tests__/resource-object-rules.test.ts
@@ -129,7 +129,7 @@ describe("resource object rules", () => {
     );
 
     test.each(["uuid", "uri", "ulid"])(
-      "passes when PATCH request body for a relationship is of the correct form (data is an collection of resource objects with id and type)",
+      "passes when PATCH request body for a relationship is of the correct form (data is a single resource object with id and type)",
       async (format) => {
         const afterJson = {
           ...baseJson,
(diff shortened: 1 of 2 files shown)


src/rulesets/rest/2022-05-25/json-api-rules/__tests__/resource-object-rules.test.ts
```diff
@@ -129,7 +129,7 @@ describe("resource object rules", () => {
     );
 
     test.each(["uuid", "uri", "ulid"])(
-      "passes when PATCH request body for a relationship is of the correct form (data is an collection of resource objects with id and type)",
+      "passes when PATCH request body for a relationship is of the correct form (data is a single resource object with id and type)",
```
jgresty, line 132: We still want a test for collections to make sure we don't regress

src/rulesets/rest/2022-05-25/json-api-rules/resource-object-rules.ts
```diff
@@ -142,6 +142,24 @@ const matchRelationshipModificationRequestData = {
   },
 };
 
+// Relationship POST, PATCH, and DELETE requests must have
+// a request body with resource objects for the relationships
+// to be added/patched/deleted.
```
jgresty, line 147: nit: this comment isn't all that helpful when it is exactly the same as the one above

jgresty (approved):
lgtm other than the missing test

src/rulesets/rest/2022-05-25/json-api-rules/__tests__/resource-object-rules.test.ts
jac-moran: oops... I probably replaced it rather than adding! Thanks!

jgresty:
> Both comments are addressed. Test re-added and green. Pls let me know any other issues @jgresty

All looks good to me, feel free to merge

PR #675, opened 2025-01-10
feat: add rule to prevent mapping objects in discriminators

jgresty:
The mapping objects type is a record of string to string, however the values
are considered references. Since the values are not deserialised as refs by
vervet, the reference objects are not copied when compiling the specs which
lead to invalid specs. This is an issue with the upstream library and is hard
to fix on our side.

The use of discriminator mapping objects should be discouraged. This patch adds
documentations to assert that as well as a linting rule to prevent mapping
objects entirely to avoid the above bug. If we fix the bug this rule may be
removed again but the docs should stay.


RaduPetreTarean (approved):
Nice!

PR #683, opened 2025-01-30
feat: allow namespaced component names

cmars:
OpenAPI generated by Typespec will use namespace-qualified component names in the generated OpenAPI, when types are in a different namespace from that of the OpenAPI spec.

This updates the componentNameCase rule to allow PascalCase component names with a dot.case namespace prefix.

~~DONE: figure out how to deal with TypeSpec materialized component names for parameters.~~

### DIFF
--- a/src/rulesets/rest/2022-05-25/specification-rules.ts
+++ b/src/rulesets/rest/2022-05-25/specification-rules.ts
@@ -1,7 +1,44 @@
 import { RuleError, Ruleset, SpecificationRule } from "@useoptic/rulesets-base";
-import { pascalCase } from "change-case";
+import { dotCase, pascalCase, snakeCase } from "change-case";
 import { links } from "../../../docs";
 import { stabilityKey } from "./constants";
+
+interface componentName {
+  localName: string;
+  localProp?: string;
+  ns?: string;
+}
+
+/**
+ * Decode the component name from a possibly namespace and property-qualified OpenAPI component name.
+ *
+ * @param name An OpenAPI component name
+ * @returns Tuple containing the local name (without namespace) and its namespace prefix (undefined if not namespaced).
+ */
+const decodeComponentName = (name: string): componentName => {
+  const parts = name.split(".");
+  if (parts.length > 1) {
+    const lastPart = parts[parts.length - 1];
+    if (lastPart === snakeCase(lastPart)) {
+      // Component is derived from a Typespec model property, of the form namespace.ModelName.property_name.
+      // This is the case for shared parameter or header components, for example.
+      return {
+        localName: parts[parts.length - 2],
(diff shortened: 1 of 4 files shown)


src/rulesets/rest/2022-05-25/specification-rules.ts
```diff
@@ -1,8 +1,45 @@
...
+ *
+ * @param name An OpenAPI component name
+ * @returns Tuple containing the local name (without namespace) and its namespace prefix (undefined if not namespaced).
+ */
+const decodeComponentName = (name: string): componentName => {
+  const parts = name.split(".");
+  if (parts.length > 1) {
+    const lastPart = parts[parts.length - 1];
+    if (lastPart === snakeCase(lastPart)) {
+      // Component is derived from a Typespec model property, of the form namespace.ModelName.property_name.
```
jgresty, line 23: Is this formally defined or just a convention? Can we rely on this if multiple services are using different typespec versions?

src/rulesets/rest/2022-05-25/specification-rules.ts
```diff
@@ -1,8 +1,45 @@
...
+      // This is the case for shared parameter or header components, for example.
+      return {
+        localName: parts[parts.length - 2],
+        localProp: lastPart,
+        ns: parts.slice(0, parts.length - 2).join("."),
+      };
+    } else {
+      // Component is derived from a Typespec model, of the form namespace.ModelName.property_name.
```
jgresty, line 31: Not sure this is accurate given the conditions needed to reach this point: length > 1 && lastPart != snakeCase

src/rulesets/rest/2022-05-25/specification-rules.ts
cmars: It's a convention. Typespec just concatenates the namespace, model name and property name with '.'.

So this is probably worth more documentation about how to name things in Typespec to make the linter happy. Should I add a section to the docs?

src/rulesets/rest/2022-05-25/specification-rules.ts
cmars: Oh, it's a comment typo. 🤦 Got it!

src/rulesets/rest/2022-05-25/specification-rules.ts
cmars: Updated the docs with an example, hopefully this makes the intent clearer.

jgresty: Thanks, I was more worried that if it is not formally defined somewhere then a change in the typespec tooling would start generating names in a different format. However looking at the way the names are generated that is unlikely

PR #687, opened 2025-02-18
chore: EXT-2825 upgrade jsonpath to fix critical vuln

acaprar:
We have a Critical sev in sweater-comb:
Screenshot 2025-02-18 at 10 49 12

Because of this, other projects are affected.

In order to fix it, we need to upgrade **jsonpath-plus** to **10.3.0**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5393,15 +5393,6 @@
         "openid-client": "^5.3.0"
       }
     },
-    "node_modules/@kubernetes/client-node/node_modules/jsonpath-plus": {
-      "version": "7.2.0",
-      "resolved": "https://registry.npmjs.org/jsonpath-plus/-/jsonpath-plus-7.2.0.tgz",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -67,7 +67,8 @@
     "yargs": "~17.6.0"
   },
   "overrides": {
-    "postman-collection": "4.2.0"
+    "postman-collection": "4.2.0",
+    "jsonpath-plus": "^10.3.0"
```
acaprar, line 71: Because we run `npm ci` in our [test pipeline](https://app.circleci.com/pipelines/github/snyk/sweater-comb/2214/workflows/35f253a5-9488-47fc-b456-5e556c432a7f/jobs/4092), package.json and package-lock.json need to be in sync other

package-lock.json
```diff
@@ -5393,15 +5393,6 @@
         "openid-client": "^5.3.0"
       }
     },
-    "node_modules/@kubernetes/client-node/node_modules/jsonpath-plus": {
-      "version": "7.2.0",
```
jgresty, line 5397: That is a big version change, are you sure this isn't going to break something?

acaprar: from the tests it seems to be fine, but if you think the tests do not cover all scenarios then we could set this version change only for `stoplight/spectral-core`

PR #736, opened 2026-04-16
fix: allow property changes when moved to polymorphic branch

fsargent:
_This pull request adds support for detecting when properties are moved to polymorphic branches (oneOf/anyOf) rather than being truly removed._

## Problem

1. Optic dereferences both the "before" (main branch) and "after" (working tree) specs
2. It walks both specs and produces a list of `IFact` objects for each property (fields) it finds
3. If a property path exists in "before" but not in "after", the differ emits a `property.removed` change event
4. Sweater-comb's rules at [property-rules.ts](../../sweater-comb/src/rulesets/rest/2022-05-25/property-rules.ts) react to these events:
    - `responsePropertyRemovalRule` (line 78-92): throws on any `property.removed`
    - `requestPropertyRemovalRule` (line 51-65): same for requests
    - `preventChangingRequestType` (line 404-425): throws when `flatSchema.type` changes
    - `preventChangingResponseType` (line 427-448): same for responses

**What happens with oneOf:**

Before: `attributes` -> `$ref: TestResultsDataAttributes` -> properties: `total_count`, `project_commit_sha`, etc.   
After: `attributes` -> `oneOf: [TestResultsDataAttributes, SecretsTestResultsDataAttributes]`

Optic sees `attributes` as a `oneOf` object now, not an `object` with direct properties. The JSON paths for the properties change from `/properties/data/properties/attributes/properties/total_count` to `/properties/data/properties/attributes/oneOf/0/properties/total_count`. The old paths are gone -> differ says "removed". But they're not removed -- they moved into a `oneOf` branch.

Optic _does_ track `oneOf`/`anyOf` paths via `polymorphicSchemas` on the `EndpointNode` (see [group-facts.ts](../../sweater-comb/node_modules/@useoptic/rulesets-base/src/rule-runner/group-facts.ts) lines 180-210), but **sweater-comb's rules never check this data**. The infrastructure to handle this exists but isn't wired up.

## Solution

The change introduces a new utility function `isPropertyMovedToPolymorphicBranch` that identifies when a property appears "removed" but actually still exists within a newly introduced oneOf/anyOf schema branch. This addresses false positives where Optic's differ incorrectly flags properties as removed when their JSON paths change due to schema restructuring.

The fix is applied to five property validation rules:

- Request property removal validation
- Response property removal validation
- Required request property addition validation
- Request property type change validation
- Response property type change validation

When a property is detected as moved to a polymorphic branch, these rules now skip throwing breaking change errors, preventing false positive violations during API evolution.

### DIFF
--- a/src/rulesets/rest/2022-05-25/property-rules.ts
+++ b/src/rulesets/rest/2022-05-25/property-rules.ts
@@ -11,6 +11,7 @@ import {
   isCompiledOperationSunsetAllowed,
   isResourceMetaProperty,
   isFullyTypedArray,
+  isPropertyMovedToPolymorphicBranch,
   specIsRemoved,
 } from "./utils";
 
@@ -55,8 +56,16 @@ const requestPropertyRemovalRule = {
     ruleContext.operation.change !== "added" &&
     !specIsRemoved(specification) &&
     !isBreakingChangeAllowed(ruleContext.custom.changeVersion.stability),
-  rule: (requestAssertions) => {
+  rule: (requestAssertions, ruleContext) => {
     requestAssertions.property.removed("not be removed", (property) => {
+      if (
+        isPropertyMovedToPolymorphicBranch(
+          property.location.jsonPath,
+          ruleContext.operation.polymorphicSchemas,
+        )
+      ) {
+        return;
+      }
       throw new RuleError({
         message: `Expected property ${property.value.key} to not be removed`,
       });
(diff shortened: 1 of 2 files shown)


fsargent:
* **#736** <a href="https://app.graphite.com/github/pr/snyk/sweater-comb/736?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snyk/sweater-comb/736?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

